PR #89, opened 2015-03-28
Misspelling fixed in RunningTests.fsx

twierzchowski:
Misspelling fixed.


kurtschelfthout:
Thanks for taking the time!

PR #104, opened 2015-06-09
Typo in ToString method of the Dec class

telekenetix:
The ToString was returning "inc" instead of "dec"


kurtschelfthout:
That was fast, thanks!

telekenetix:
No problem! Glad to contribute, even if it was something tiny.

PR #112, opened 2015-07-05
Workarounds for non portable code

mstyura:
Implemented quick and dirty workarounds for `System.Console`, `CultureInfo.GetCultures` which can't be used in `Portable259`.
Adjusted `AutoGenerateBindingRedirects` and `Reference.Private (CopyLocal)` according to recommendations: [Notes and Guidance on FSharp.Core](https://fsharp.github.io/2015/04/18/fsharp-core-notes.html). 
Unfortunately compiler bug ([Inlining breaks PCL-defined methods in normal assemblies](http://visualfsharp.codeplex.com/workitem/144)) is presented in `divMod` and `divMod64` even when `--nooptimizationdata` is specified.
UPD: Re-read bug compiler bug description and it seems like `--nooptimizationdata` only helps for functions which is not marked as `inline`.

### DIFF
--- a/docs/csharp/App.config
+++ b/docs/csharp/App.config
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
-    </startup>
+  <startup> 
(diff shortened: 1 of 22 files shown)


kurtschelfthout:
Ah interesting on the inlining bug. Is that why the appveyor build for this PR fails some tests with MissingMethodException for Math.Sign?

mstyura:
@kurtschelfthout yes, if you remove `inline` from `divMod` all tests are passing.

Issue #120, opened 2015-07-24
Conflict with operator for Gen


paulyoung:
I have some tests that use custom operators, in particular `<*>`, which conflicts with the operator defined here: https://github.com/fscheck/FsCheck/blob/15b61f4a463691f7363b3dc71f022514c556867c/src/FsCheck/Gen.fs#L471

I followed [the advice of Tomas Petricek](http://stackoverflow.com/questions/19682432/global-operator-overloading-in-f/19683562#19683562) and defined mine as static members.

Do you know if doing the same would prevent the operator from `Gen` taking precedence? If so would you be open to a pull request that makes this change?

kurtschelfthout:
Sure, if it works.

kurtschelfthout:
Yes, 2.0.5 is out. If you get antsy, there is a NuGet feed on appveyor of every build. See the readme.

PR #122, opened 2015-07-24
Define custom operators for `Gen` as static members.

paulyoung:
This pull request defines custom operators for `Gen` as static members since definitions using `let` conflict with user-defined custom operators, as described in #120.


kurtschelfthout:
Thanks!

paulyoung:
Will this make it into a release in the near future?

kurtschelfthout:
this is in 2.0.5;  if you get antsy, there is a nuget feed associated with the appveyor builds, see the readme.

PR #129, opened 2015-08-02
remove trailing whitespace from FsCheckAddin.fs

sideeffffect:


ploeh:
Looks good to me, but I'll let @kurtschelfthout make the decision whether or not to pull :+1:

PR #133, opened 2015-08-12
Minimal failing case for mono stackoverflow bug

adbrowne:
This is obviously not ready to merge. I'm more posting this for discussion.

We have been having some problems with large tests causing a stackoverflow under mono. Our tests pass on the Microsoft CLR. I presume this is due to mono not handling tail call optimisation in all the cases where the CLR does.

This tests fails under mono (tested with: Mono JIT compiler version 4.0.2 (Stable 4.0.2.5/c99aa0c Wed Jun 24 10:04:37 UTC 2015).

Your recent change to remove delay means that the stackoverflow happens in less places. Before that I didn't need the Gen.constant 1 to cause the issue.

Do you have any advice on how we might go about fixing this?

They async computation expression goes to a lot of work to use a trampoline and I wonder if this might be why?
https://github.com/fsharp/fsharp/blob/master/src/fsharp/FSharp.Core/control.fs

### DIFF
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -253,3 +253,15 @@ module Gen =
         |> Seq.pairwise 
         |> Seq.forall obj.ReferenceEquals
         |> Assert.True
+    [<Property>]
+    // this currently fails on mono
+    let ``recursive generator should not stackoverflow`` () =
+        let rec recursiveGenerator (iterations : int) : Gen<unit> = gen {
+            if iterations < 100000 then
+                let! _ = Gen.constant 1
+                return! recursiveGenerator (iterations + 1)
+            else
+                return ()
+        }
+
+        Prop.forAll (recursiveGenerator 0 |> Arb.fromGen) (fun () -> true)


adbrowne:
Looks like the Travis build failed with a stack overflow as expected.

Also I have updated a mono ticket with a simple failing case for mono: https://bugzilla.xamarin.com/show_bug.cgi?id=12635. However I think mono not supporting tail calls completely has been known for some years so I don't know if that will be solved soon.

adbrowne:
Please ignore that mono ticket. I don't think that bug report is actually correct. I will try and create a minimal failing case so we understand what doesn't work on mono that does under the Microsoft CLR.

kurtschelfthout:
Yes - the fact that recursive generators consume stack space is known (to me at least). This is because, essentially, the function inside `Gen`, i.e. `size:int rand:StdGen -> value:'a` itself becomes recursive although the computation expression itself looks tail recursive.

In your example the generator expands to something like (ignoring the `if`):

    gen.Delay(fun () -> gen.Bind(Gen.constant 1, fun k -> gen.Return(recursiveCall ())))

which essentially becomes

    Gen (fun n r -> Gen (fun _ _ -> 1) >>= (fun k -> fun _ _ -> recursiveCall()))

And clearly the whole thing nests, i.e. you can replace `recursiveCall` with the entire expression and nothing is in tail position.

This is a long winded way of saying that as it currently stands, recursive generators don't really scale in FsCheck. I do believe you are correct on the trampolining - a full and proper solution would require using continuation passing style for the `Gen` type and I think trampolining - although using continuation passing style (a change that should be invisible to the user) would already help a lot.

So I believe that the difference between Mono and the CLR is just that the CLR may have more stack space or uses it more sparingly than Mono, or the CLR JIT somehow optimizes things better. I don't think it's directly related to tail recursion, in that the whole thing is not tail recursive. Though it you have evidence to the contrary, let me know. And it's fascinating that the CLR is pretty good at optimizing this - would love to understand more - though you'd still hit it on the CLR in more elaborate cases.

Another thing - in the latest release I have put the delay back, since its removal messed up other things.

That doesn't mean there isn't anything you can do right now though. Try to avoid the use of recursive computation expressions by using `Gen.sequence` which does not consume stack space - maybe we need a few more combinators in the same vein, e.g. `Gen.unfold` or `Gen.fold`. Not sure.

More straightforwardly, the `gen` computation expression supports `while` and `for`, neither of which should consume stack space, and rewriting recursion to use those should be relatively straightforward.

adbrowne:
Thanks Kurt, a combination of converting to using a while loop and limiting the EndSize is letting me run on mono without crashing. In this case we are using Gen to simulate running programs in parallel by threading the Gen calls through our interpreter so there is a lot of recursion that is difficult to avoid.

Now that I've dug in a little bit more I'm very impressed that the CLR seems to be able to use stack space so sparingly. I've tried checking the stack in the debugger while the tests are running and it seems to remain very small.

I have been experimenting with a version of Gen that uses CPS and a trampoline - but have not yet had much success. All the existing F# trampolines I've found cause a stack overflow on mono. I am now looking at some of the scala and C# implementations for inspiration.

kurtschelfthout:
There is a branch Gen-CPS with a tail-recursive bind that you may want to try. One or more tests still fail, but it's worth trying, think that's due to another issue.

PR #134, opened 2015-08-18
Fixes compatiblity issues with VB.NET

pblasucci:
Currently, the VB compiler does NOT recognize any extension methods defined in FsCheck. This causes serious awkwardness as the vast majority of the non-F# APIs reply on extensions. Fortunately, the fix is simple. More details maybe found here: http://latkin.org/blog/2014/04/30/f-extension-methods-in-roslyn/


kurtschelfthout:
Brilliant. Thank you!

Issue #137, opened 2015-08-22
Convert a test in a properties based test in C#


gsscoder:
Given an XUnit test like this:
```csharp
[Theory]
[InlineData(new[] {"-123"}, -123L)]
[InlineData(new[] { "-1" }, -1L)]
[InlineData(new[] { "-9223372036854775807" }, -9223372036854775807)] // long.MaxValue * -1
public void Parse_negative_long_value(string[] arguments, long expected) {
  // Fixture setup in attributes
  // Exercize system 
  var result = InvokeBuild<Simple_Options>(arguments);
  // Verify outcome
  ((Parsed<Simple_Options>)result).Value.LongValue.ShouldBeEquivalentTo(expected);
  // Teardown
}
```

What I'd like to do is remove ``InlineData`` attribute and let **FsCheck** generate random data for me.

I've tried using `Prop.ForAll` and `Property` attribute, but still experiencing some difficulties.

Full test class source code: https://github.com/gsscoder/commandline/blob/master/tests/CommandLine.Tests/Unit/Core/InstanceBuilderTests.cs#L62.

Maybe could be a simple solution add a new **F#** to the solution, explicitly targeted to properties based testing?

Thanks in advance and thanks for sharing this great tool.

moodmosaic:
One way of doing that, is to create a new F# Class Library project, open the Package Manager Console in Visual Studio and execute:

```
install-package FsCheck.Xunit
```

Instead of `Theory` and `InlineData` just use `Property` to have FsCheck generate test cases for your test (or rather, *property*):

```f#
module Tests

open FsCheck.Xunit

[<Property>]
let ``Parse negative long value`` (arguments : string array) (expected : int) =
    // Define your test/property here.
    ()
```

kurtschelfthout:
Pretty much everyhing that works in F# should work in C#. Personally I would use F# but really as far as FsCheck is concerned there are only syntactic differences. So it's your call.

I can't really comment specifically unless you tell me more about the actual problems.

The one shift you should make is that you now just encode a list of inputs with their corresponding expected outputs. With FsCheck, you should figure out the properties that hold based on random inputs, i.e. if you're going to just replace Theory and InlineAttribute with Property, then clearly things won't work because FsCheck will generate some random arguments, and some random and completely independent expected results, and so your test will fail unless you are _very_ lucky :)

[...]

PR #142, opened 2015-09-01
Added information about versioning to README.md.

ploeh:
When troubleshooting versioning problems (e.g. with NuGet), it can be
useful for users to know that they can rely on a particular library to
follow Semantic Versioning. This enables users to assume that there are no
breaking changes among versions sharing the same major version, but that
there *are* breaking changes between versions with two different major
versions.

Case in point: https://github.com/fscheck/FsCheck/issues/137#issuecomment-136592290

In this pull request, I've implicitly *assumed* that FsCheck follows Semantic Versioning, which I don't know for certain, but that I'm seeing quite a few signs of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ Releases:
 * [FsCheck with xUnit.NET integration](http://nuget.org/List/Packages/FsCheck.Xunit)
 * [FsCheck with NUnit integration](http://www.nuget.org/packages/FsCheck.Nunit/)
  
+FsCheck follows [Semantic Versioning 2.0.0](http://semver.org/spec/v2.0.0.html).
+
 All AppVeyor builds are available using the NuGet feed: https://ci.appveyor.com/nuget/fscheck
 
 If using Paket, add the source at the top of `paket.dependencies`.


kurtschelfthout:
Yes, though I've recently noticed I've broken some of the rules out of ignorance, by adding (backwards compatible) features to patch releases. So I guess haven't strayed too far from the flock, but I'll try to avoid in the future.

I've wondered what to do about the upgrade to xunit2/Nunit 3 though. I've tentatively settled on minor version increase, because FsCheck's API does not change, except for some obsoleted methods. semver.org seems to give the same indication in its FAQ.

ploeh:
The question is whether a client would be able to update e.g. *FsCheck.Xunit* and still be able to compile all code.

If one uses only *FsCheck.Xunit*, one probably can, but we can't guarantee it.

Imagine a client using *FsCheck.Xunit*. Apart from using the `[<Property>]` attribute, such a client may be mixing properties with Parameterized Tests, using the `[<Theory>]` attribute. In xUnit.net 1, this attribute is defined in the `Xunit.Extensions` namespace, but in xUnit.net 2, it's been moved to the `Xunit` namespace.

Such a client would most likely have an `import` of `Xunit.Extensions`, and that will not compile with xUnit.net 2.

Therefore, strictly speaking, the upgrade is a breaking change.

It's quite a pain. We're having the same problem with AutoFixture, because of its xUnit.net Glue Library, and the way we got around it over there was to keep *AutoFixture.Xunit* as is, and release a new library called *AutoFixture.Xunit2*. After some time has passed, we can deprecate *AutoFixture.Xunit*.

In retrospect, I've begun to wonder whether the decision to keep the core AutoFixture library, and all its satellite libraries in sync, version-wise, was a good idea. I don't think it was.

As far as I can tell, FsCheck is in the same situation.

kurtschelfthout:
Sure, but that breakage is already documented by upgrading from xunit 1.9.2 to xunit 2.0. The fact that FsCheck essentially forces you to do that is technically of no consequence. 

Taken from the FAQ:

What should I do if I update my own dependencies without changing the public API?

That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it's obviously a minor level increment.

ploeh:
The most important question, I hope, is how this impacts users.

I know that I'm being difficult here, but bear with me for a moment. Personally, I don't much mind in this particular situation, since I already know what's going on. New users, on the other hand, do not, and I'm arguing in their place.

As a user, I use Semantic Versioning to answer questions such as: *if I update this library, will it break my code?*

If I didn't already know *FsCheck.Xunit*, and saw that the version went from 2.0.6 to 2.1.0, I'd conclude that I could safely upgrade *FsCheck.Xunit* without any problems. Because of the breaking changes in xUnit.net, this might then turn out to be false.

I admit that the quoted FAQ seems to point in a different direction, but I don't find it helpful. Ultimately, you'll need to decide whether following Semantic Versioning is a convenience to the *FsCheck* developers, or its users.

Sometimes, when writing and maintaining a reusable library, we developers must take on extra pain, in order to prevent that pain from impacting all the library's users. It's a question of scalability. A hit we take may make work easier for thousands of other developers. A mistake baked into a library can impact thousands of developers.

For the record, the solution with *AutoFixture.Xunit2* that we have in AutoFixture isn't flawless either. Most people figure it out on their own, but we do occasionally get [questions from people who *don't* figure it out](https://github.com/AutoFixture/AutoFixture/issues/436).

There's no clear solution, so I'm happy with any decision you might make. I only wanted to point out that pretending that this isn't a breaking change may not be helpful to users.

kurtschelfthout:
You're not being difficult, we're talking through it. It's helpful.

I don't feel like we would be pretending this isn't a breaking change - in fact it isn't a breaking change in FsCheck's API. By updating minor version, we're being honest about where the breaking change is: in xunit, not FsCheck or FsCheck.Xunit.

In fact, if folks are using FsCheck.Xunit and xunit at this point they almost certainly have some restriction like xunit < 2.0 in their projects (for paket users, actually not sure if you can make it explicit in nuget). In other words, when upgrading to the new FsCheck.Xunit (which would require xunit >= 2.0) the problem should be flagged and it's clear what the impact is.

Another flip side of doing a major version increase is that users of just FsCheck may hold off on upgrading thinking there are breaking changes, when there aren't any. So now we're lying to them. Though I admit this is really a distinct problem, viz. that we're versioning all 3 packages with the same version number, and I don't have the same qualms about increasing FsCheck.Xunit's version when really there hasn't been a change to it.

Overall, I'm starting to regret getting into the business of integrating with test runners more tightly. In my own projects, lately I've switched to just using `QuickCheckThrowOnFailure`.

ploeh:
> this is really a distinct problem, viz. that we're versioning all 3 packages with the same version number

Agreed. As I wrote, I have the same problem in AutoFixture, and I think it was a mistake to version it like that, but now I feel that I can't change the rule - at least not in AutoFixture 3, as it would confuse too many people.

Likewise, I think FsCheck would need to stick to the versioning strategy already established, so I agree that incrementing the version to FsCheck 3 only because of the change to FsCheck.Xunit, isn't desirable.

> Overall, I'm starting to regret getting into the business of integrating with test runners more tightly.

Don't regret that; I'm getting much benefit out of *FsCheck.Xunit*, and I try to help all I can with it :smile: 

Bottom line: while not optimal, I get the sense that you're leaning towards releasing this as FsCheck 2.1.0, and I support that decision.

kurtschelfthout:
Well, glad you're getting benefit from it, personally not sure if it's worth all the complexity, but of course it's a personal view - after all I'm getting the bulk of the complexity :) . All that said, I have no plans to discontinue support whatsoever, and certainly appreciate the help.

Issue #86, opened 2015-02-18
support xUnit 2


ctaggart:
It would be great if you could add support for xUnit 2 http://xunit.github.io/. May be call the package `FsCheck.Xunit2` if you wish to continue support xUnit 1.

kurtschelfthout:
You mean essentially changing xunit dependency version in Fscheck.Xunit package and project, right? Because it already works with the xunit 2.0 runner as far as I can tell.

ctaggart:
Correct. Our projects will not build with xunit 2 and FsCheck.Xunit as dependencies.

kurtschelfthout:
I'm sympathetic, but given that they are still in prerelease am hesitant to update FsCheck.Xunit. Maybe FsCheck.Xunit2 is a better option. Have my hands full with the C# revamp at the moment though.

MartinJohns:
I just wanted to note that Xunit is not in a pre-release state anymore. :-)

kurtschelfthout:
Yes…noted ;)

jhamm:
I have been working on upgrading XUnit2. There are only a few issues left to work out.

jhamm:
Created PR xunit/xunit#403 to fix the handling of attribute parameters that are array types.

kurtschelfthout:
:+1:

kurtschelfthout:
@jhamm do you think we can put this in the upcoming FsCheck 2.0 release? Any issues on the FsCheck side I can help with?

jhamm:
Here is the work I have so far jhamm/FsCheck branch support-xunit2.

If everything is setup, this will run 123 tests, based on https://ci.appveyor.com/project/kurtschelfthout/fscheck, the same number of tests are executed, so that is a good thing.
```
 packages\xunit.runner.console\tools\xunit.console.exe tests\FsCheck.Test\bin\debug\fscheck.test.dll
```

These are the remaining issues that I can see:

- If you get the error "System.Runtime.Serialization.SerializationException : Unable to find assembly 'FsCheck.Xunit, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null'."
    - I temporarily fixed this by copying FsCheck.Xunit to the directory "packages\xunit.runner.console\tools"
    - If you get this error when using the VisualStudio Runner, copy FsCheck.Xunit.dll to %TEMP%\VisualStudioTestExplorerExtensions\xunit.runner.visualstudio.2.0.0\build\_common.

[...]

PR #145, opened 2015-09-05
xunit2 - add QuietOnSuccess support back in.

jhamm:


ploeh:
Looks good to me (but I admit that I haven't tested it yet) :+1:

moodmosaic:
On Visual Studio 2015 with TestDriven.Net 3.9 Beta 3, specifying `[<Property(QuietOnSuccess = true)>]` still doesn't suppress the output in case of success. :confused:

This works fine with FsCheck.Xunit 2.0.7.

--

**How I built the NuGet Packages locally**

[SourceLink](https://chocolatey.org/packages/SourceLink) contains `pdbstr.exe` which is [required by SourceLink.Fake](http://ctaggart.github.io/SourceLink/fake.html), so I installed it via chocolatey:

```
choco install SourceLink
```

After that, I run a build to create the NuGet Packages:

```
build.cmd nuget
```

Then, I added the `bin` folder as a Package Source and installed FsCheck.Xunit from there.

jhamm:
I downloaded and installed TestDriven.net. I am making progress on this. I see part of the issue, I was passing the success message to the xunit2 message bus and this was writing the message to the output window.
I have the tests that return void/unit working. The tests that return a value are not working because they are writing to the output the value (in the case of the FsCheck.XUnit.CSharpExamples BCL test method, this is returning "True") and causing the output window to pop up.

jhamm:
I have checked in some more changes that remove the "Ok, passed 100 tests" message when this flag is set.

I would like some help trying to narrow down what the issue is with the FsCheck.XUnit.CSharpExamples test named Bcl. I have decorated this with the QuietOnSuccess = true attribute property and the output of "True" continues to cause the output window to open.

The Bcl2 test returns void and does not cause the output window to open.

At this time I do not see where this is being written to the output.
Any help will be appreciated. Thanks!

moodmosaic:
I pulled locally the latest from [jhamm:xunit2](https://github.com/fscheck/FsCheck/compare/xunit2...jhamm:xunit2), run the following property, but the output looks incorrect now:

```fsharp
[<Property(QuietOnSuccess = true)>]
let revUnit (x:char) = 
    List.rev [x] = [x]
```

Output:

```
0 Passed, 0 Failed, 0 Skipped
```

--

I'd expect the output to be:

```
1 Passed, 0 Failed, 0 Skipped
```

jhamm:
I can now reproduce this behavior, it took a while because I am not familiar with TestDriven.net. I will look at a solution to this.

jhamm:
When testing I only used the C# test, thinking that this was the edge case and the F# test would work the same.
I have both C# and F# FsCheck Property tests working now. The trick was to return an empty string.

examples/FsCheck.XUnit.CSharpExamples/ReverseFixture.cs
```diff
@@ -6,17 +6,25 @@ namespace FsCheck.XUnit.CSharpExamples
...
-
-        [Property]
+        [Property(QuietOnSuccess = true)]
         public bool Bcl(int[] xs)
         {
           return xs.Reverse().Reverse().SequenceEqual(xs);
         }
 
+        [Property(QuietOnSuccess = true)]
+        public void Bcl2(int[] xs)
```
moodmosaic: Can't seem to easily figure out the kind of feedback this test provides. Is it really needed?

moodmosaic:
While I haven't investigated whether returning an empty string can the best approach to solve this, I can now confirm that **it now works fine on my machine** :+1:

---

As a side note: You may want to consider [keeping the git history cleaner](http://mat-mcloughlin.net/2015/06/27/keeping-a-clean-git-history/) now that you've had to add a bunch of commits to address this.

examples/FsCheck.XUnit.CSharpExamples/ReverseFixture.cs
jhamm: I was using it to determine what the C# void return would do with TestDriven.NET. If you think I should remove it, I can do that.

jhamm:
Thanks for the tip about `git rebase -I`. After the command `git rebase` I had to use `git push origin --force` to get the changes up to GitHub. Is this the correct thing to do in this situation? I think it may be ok because this pull request is only being worked on by one person.

examples/FsCheck.XUnit.CSharpExamples/ReverseFixture.cs
moodmosaic: Other than helping in debugging, it doesn't seem to provide any sort of *automated* feedback, unless I miss something.

moodmosaic:
The commit log is much cleaner now :+1:

examples/FsCheck.XUnit.CSharpExamples/ReverseFixture.cs
ploeh: FWIW, when I originally developed the *QuietOnSuccess* feature, I added a similar 'test' to the F# tests. There was no automated feedback (because, in general, there was no test coverage of FsCheck.Xunit), but I used the test for the documentation to ensure that it at least compiles. I think that 'test' is still around.

ploeh:
> Is this the correct thing to do in this situation?

There are advantages and disadvantages to doing that. In general, a pull request branch or fork only has a single author, so it's okay to force push to it. (It's still absolutely not okay to force push to *master*, or other shared branches, but here in FsCheck, that point is moot, since you don't have commit access to *master*.)

The advantage of rebasing a pull request branch is that the commit history becomes cleaner. This mainly helps when one is doing commit history archaeology in the future (digging through commits in order to figure out why things are as they are; a current example was when I was trying to figure out [why *QuietOnSuccess* no longer works](https://github.com/fscheck/FsCheck/issues/86#issuecomment-137949096)).

The disadvantage is that *during pull request review*, force pushing essentially resets the entire branch, which means that it's difficult for the reviewer to identify what changed since the last review pass. That's not a big problem for small pull requests, but can be for large pull requests.

The *avoid thrashing* advice was the most controversial of my [10 tips for better pull requests](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests), and the only one I'm beginning to regret that I gave.

ploeh:
Just tested with my usual test project: all looks good so far :+1: 

---

# Details

## NuGet
```
PM> Install-Package FsCheck.Xunit -Source https://ci.appveyor.com/nuget/fscheck -Version 2.1.0-xunit2-b307
```

```
PM> Get-Package

Id                                  Versions               
--                                  --------               
FsCheck                             {2.1.0-xunit2-b307}    
FsCheck.Xunit                       {2.1.0-xunit2-b307}    
FSharp.Core                         {3.1.2.5}              
Unquote                             {3.0.0}                
xunit.abstractions                  {2.0.0}                
xunit.extensibility.core            {2.1.0-beta4-build3109}
xunit.extensibility.execution       {2.1.0-beta4-build3109}
```

## QuietOnSuccess

### All suppressed

```
------ Test started: Assembly: ConwaysGameOfLife04.dll ------

10 passed, 0 failed, 0 skipped, took 1,28 seconds (xUnit.net 2.1.0 build 3109).
```

### All but one suppressed

```
------ Test started: Assembly: ConwaysGameOfLife04.dll ------

Output from Ploeh.Katas.GameOfLifeProperties.Neighbors don't include original cell:
  Ok, passed 100 tests.

10 passed, 0 failed, 0 skipped, took 1,30 seconds (xUnit.net 2.1.0 build 3109).
```

## Stack traces on exceptions

```
------ Test started: Assembly: ConwaysGameOfLife04.dll ------

Test 'Ploeh.Katas.GameOfLifeProperties.Non-neighboring live cells don't influence cell' failed: FsCheck.Xunit.PropertyFailedException : 
Falsifiable, after 26 tests (64 shrinks) (StdGen (1076175853,296055324)):
Original:
(-8, 8,
 [(-10, 6); (-1, -12); (-9, 7); (0, -11); (-8, 8); (1, -10); (-7, 9); (2, -9);
  (5, 10); (-3, -8); (6, 11); (-2, -7); (7, 12); (-1, -6)])
Shrunk:
(0, 0, [(0, 1); (0, 0); (1, 0)])

---- Swensen.Unquote.AssertionFailedException : Test failed:

Dead = Alive
false

	
	----- Inner Stack Trace -----
	C:\Users\mark\Documents\Katas\ConwaysGameOfLife04\ConwaysGameOfLife04\Properties.fs(102,0): at Ploeh.Katas.GameOfLifeProperties.Non-neighboring live cells don't influence cell(Int32 cell_0, Int32 cell_1, FSharpList`1 otherCells)
	--- End of stack trace from previous location where exception was thrown ---
	at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
	at FsCheck.Runner.invokeAndThrowInner@306-1.Invoke(Object[] o)
	at <StartupCode$FSharp-Core>.$Reflect.Invoke@958-4.Invoke(T1 inp)
	at FsCheck.Testable.evaluate[a,b](FSharpFunc`2 body, a a)

Output from Ploeh.Katas.GameOfLifeProperties.Non-neighboring live cells don't influence cell:
  Falsifiable, after 26 tests (64 shrinks) (StdGen (1076175853,296055324)):
  Original:
  (-8, 8,
 [(-10, 6); (-1, -12); (-9, 7); (0, -11); (-8, 8); (1, -10); (-7, 9); (2, -9);
  (5, 10); (-3, -8); (6, 11); (-2, -7); (7, 12); (-1, -6)])
  Shrunk:
  (0, 0, [(0, 1); (0, 0); (1, 0)])

9 passed, 1 failed, 0 skipped, took 1,83 seconds (xUnit.net 2.1.0 build 3109).
```

ploeh:
This pull request looks good to me :+1:

kurtschelfthout:
Thanks everyone.

I feel compelled to remark that while I'm not really obsessed with keeping PRs clean, it does help with reviewing and finding bugs after the fact. And clearly @ploeh's PRs in particular are a thing of beauty.  :)

jhamm:
I really appreciate the time everyone has taken to help. Thanks!

PR #146, opened 2015-09-10
update xunit2 references to 2.1.0-rc1

jhamm:
The nuget packages.config files were not included like in a3ba95165799f836a1919ec4c206d52533002d40.


kurtschelfthout:
Thanks! Do you know if they have an ETA for the final 2.1?

jhamm:
From what I can find, there is no hard date. Only information I can find is that it should ship before DNX goes RTM.

PR #153, opened 2015-09-28
add documentation for installing with Paket

sideeffffect:


.travis.yml
```diff
@@ -4,9 +4,10 @@ before_script:
  - chmod +x build.sh
 
 script: 
-  - ./build.sh RunTests
+  - ./build.sh GenerateDocs
```
kurtschelfthout: Can you move this change to a separate PR please. I will be happy to accept it but I'd like it in a separate commit.

README.md
```diff
@@ -39,6 +42,19 @@ nuget FsCheck 2.0.4
...
+Add reference to the following assemblies in the apropriate project(s):
+
+```
+packages/NUnit.Runners/tools/lib/nunit.core.dll
+packages/NUnit.Runners/tools/lib/nunit.core.interfaces.dll
+```
+
+Also make sure your project(s) has target `.NET 4.5` or greater.
+
+
```
kurtschelfthout: This doesn't belong in the readme, can you add it to https://fscheck.github.io/FsCheck/RunningTests.html "Using FsCheck.NUnit" paragraph instead?

docs/content/QuickStart.fsx
```diff
@@ -10,7 +10,20 @@ The fastest way to understand how FsCheck works is by writing some *properties*
...
+([Paket doesn't run any powershell scripts](https://fsprojects.github.io/Paket/faq.html#Does-Paket-run-install-ps1-scripts)
+that usually do this for you):
+
+```
+packages/NUnit.Runners/tools/lib/nunit.core.dll
+packages/NUnit.Runners/tools/lib/nunit.core.interfaces.dll
+```
+
+## Usage
+Open an fsx file and start with:*)
```
kurtschelfthout: Don't feel like this belongs here either, the quickstart uses vanilla FsCheck with a shallow reference to xUnit integration at the end. It would be better to extend the xunit section somewhat to basically merely acknowledge the existence of FsCheck.NUnit, but just put a link to the more NUnit section I mentioned earlier.

docs/tools/generate.fsx
```diff
@@ -22,6 +22,7 @@ let info =
 // For typical project, no changes are needed below
 // --------------------------------------------------------------------------------------
 
+#I "../../packages/FAKE/tools/"
```
kurtschelfthout: I don't think this is necessary.

docs/tools/generate.ja.fsx
```diff
@@ -22,6 +22,7 @@ let info =
 // For typical project, no changes are needed below
 // --------------------------------------------------------------------------------------
 
+#I "../../packages/FAKE/tools/"
```
kurtschelfthout: I don't think this is necessary.

.travis.yml
kurtschelfthout: Now that I think about it, the target should be "CI". That's the one AppVeyor uses.

docs/tools/generate.fsx
sideeffffect: actually this is needed for `Microsoft.Web.XmlTransform.dll`

kurtschelfthout: Sorry, I don't understand. Please open a separate PR describing what the problem is you're trying to solve, and how this solves it.

kurtschelfthout:
Brilliant, thanks.

Issue #160, opened 2015-10-04
Assert FsCheck with Unquote instead of xunit.assert


ploeh:
The *FsCheck.Test* project has a reference to [xunit.assert](https://www.nuget.org/packages/xunit.assert), and uses the xUnit.net assertion library to express various assertions.

xunit.assert isn't particularly F#-friendly, and has some (in my opinion) unfortunate implicit behaviour; [here's an example](https://xunit.codeplex.com/workitem/9856) - even though that particular issue was marked as *closed*, it was never properly addressed.

When expressing assertions in F#, I find [Unquote](https://www.nuget.org/packages/Unquote) much more suitable. Would it be OK to remove xunit.assert and rewrite the assertions with Unquote?

There are some 34 assertions that currently depend on xunit.assert. I'll be happy to do this work in a pull request, if you agree that this would be a good idea.

ploeh:
A couple of questions regarding this has come up:

I can't figure out exactly how *FsCheck.Test* references *xunit.assert*. If I ask the NuGet package manager console to report the packages for the project, I only get this:

```
PM> Get-Package -ProjectName FsCheck.Test

Id                                  Versions                                 ProjectName
--                                  --------                                 -----------
xunit.runner.visualstudio           {2.1.0}                                  FsCheck.Test
```

Still, I can see in the project's references that *xunit.assert* has this `HintPath`: `..\..\packages\xunit.assert\lib\portable-net45+win8+wp8+wpa81\xunit.assert.dll`, so it looks like it's a package of some sort.

It's not included in `paket.references` either, so it's not clear to me exactly how this dependency is managed.

Still, I attempted to remove the dependency simply by deleting the reference from Visual Studio. As expected, this causes *FsCheck.Test* to fail compilation, but it also causes the following projects to fail compilation:

- FsCheck.CSharpExamples
- FsCheck.MsTest.Examples
- FsCheck.NUnit.CSharpExamples

Why does this happen?

As far as I can tell, none of these projects depend on *FsCheck.Test*, so it's not clear to me why this happens.

kurtschelfthout:
I'd love to convert to unquote. 

The behaviour you describe is magical. Though I can explain why nuget only finds the one package: paket has to do all kinds of special stuff for it because it is a vs test runner. Don't expect nuget to see the others.

ploeh:
What's the correct way to remove the *xunit.assert* reference, and add the *Unquote* reference?

Why does compilation fail for those unrelated projects?

[...]

PR #162, opened 2015-10-05
Migrate from xunit assertions to unquote

ploeh:
Addresses #160 

A couple of comments:

- There's more than one way to express many of the assertions with Unquote. As you can tell, I tend to favour the built-in operators `=!` and `<>!` over `test <@ ... @>` in most cases. It would, however, also be possible to stick more consistently with the `test <@ ... @>` style. Let me know if I should change that.
- With the built-in operators `=!` and `<>!`, the assertions become so terse that they almost 'drown' in the test. If we stick with these operators, I would recommend adopting this heuristic for formatting tests: http://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,6 +3,7 @@ source https://www.nuget.org/api/v2
 nuget FAKE
 nuget NUnit
 nuget NUnit.Runners
+nuget Unquote
 nuget xunit
 nuget xunit.runner.console
 nuget xunit.runner.visualstudio version_in_path: true
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -5,6 +5,7 @@ nuget NUnit
 nuget NUnit.Runners
 nuget Unquote
 nuget xunit
+nuget xunit.core
 nuget xunit.runner.console
 nuget xunit.runner.visualstudio version_in_path: true
 nuget NuGet.CommandLine
--- a/paket.lock
+++ b/paket.lock
@@ -15,124 +15,140 @@ NUGET
     System.Collections (4.0.10) - framework: dnxcore50
       System.Diagnostics.Debug (>= 4.0.0) - framework: dnxcore50
       System.Resources.ResourceManager (>= 4.0.0) - framework: dnxcore50
-      System.Runtime (>= 4.0.0) - framework: dnxcore50
+      System.Runtime (>= 4.0.0)
       System.Runtime (>= 4.0.20) - framework: dnxcore50
       System.Runtime.Extensions (>= 4.0.0) - framework: dnxcore50
       System.Threading (>= 4.0.0) - framework: dnxcore50
     System.Diagnostics.Contracts (4.0.0) - framework: dnxcore50
-      System.Runtime (>= 4.0.0) - framework: dnxcore50
+      System.Runtime (>= 4.0.0)
     System.Diagnostics.Debug (4.0.10) - framework: dnxcore50
-      System.Runtime (>= 4.0.0) - framework: dnxcore50
+      System.Runtime (>= 4.0.0)
     System.Globalization (4.0.10) - framework: dnxcore50
-      System.Runtime (>= 4.0.0) - framework: dnxcore50
+      System.Runtime (>= 4.0.0)
     System.IO (4.0.10) - framework: dnxcore50
       System.Globalization (>= 4.0.0) - framework: dnxcore50
-      System.Runtime (>= 4.0.20) - framework: dnxcore50
-      System.Text.Encoding (>= 4.0.0) - framework: dnxcore50
+      System.Runtime (>= 4.0.20)
+      System.Text.Encoding (>= 4.0.0)
       System.Text.Encoding (>= 4.0.10) - framework: dnxcore50
       System.Text.Encoding.Extensions (>= 4.0.0) - framework: dnxcore50
       System.Threading (>= 4.0.0) - framework: dnxcore50
-      System.Threading.Tasks (>= 4.0.0) - framework: dnxcore50
+      System.Threading.Tasks (>= 4.0.0)
     System.Linq (4.0.0) - framework: dnxcore50
-      System.Collections (>= 4.0.10) - framework: dnxcore50
-      System.Diagnostics.Debug (>= 4.0.10) - framework: dnxcore50
-      System.Resources.ResourceManager (>= 4.0.0) - framework: dnxcore50
-      System.Runtime (>= 4.0.20) - framework: dnxcore50
-      System.Runtime.Extensions (>= 4.0.10) - framework: dnxcore50
+      System.Collections (>= 4.0.10)
+      System.Diagnostics.Debug (>= 4.0.10)
+      System.Resources.ResourceManager (>= 4.0.0)
+      System.Runtime (>= 4.0.20)
+      System.Runtime.Extensions (>= 4.0.10)
+    System.Linq.Expressions (4.0.10) - framework: dnx451, dnxcore50, monoandroid, monotouch, xamarinios, winv4.5, wpv8.0, portable-net45+win80+wp80+wpa81, >= net45
     System.ObjectModel (4.0.10) - framework: dnxcore50
-      System.Collections (>= 4.0.10) - framework: dnxcore50
-      System.Diagnostics.Debug (>= 4.0.10) - framework: dnxcore50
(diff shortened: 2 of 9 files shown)


kurtschelfthout:
Nice work, I have no problems with the style whatsoever. 

Failing on Appveyor and working on travis though, that's a first. It looks like on AppVeyor it's having trouble loading the FsCheck dll - likely a side-effect of something that Unquote tries to do? Do you see a pattern in the 10 failing tests?

ploeh:
It *almost* looks like that the failing tests are all the tests that use quoted expressions (the `test <@ ... @>` style); the only exception is *FsCheck.Test.Gen.should generate functions from nullable values*. That test is a bit special in the way `test` is used, so I wonder if it would also fail if I rewrote it to a more complete quoted expression...

That's as much as I've been able to figure out so far. It's getting late, so that's how far I'll get today.

ploeh:
To clarify: *FsCheck.Test.Gen.should generate functions from nullable values.* **passes** even though it uses a quoted expression, which makes it fall outside the pattern I thought I was seeing.

ploeh:
Here's a theory that could explain why *FsCheck.Test.Gen.should generate functions from nullable values* **passes** while all the other properties using `test <@ ... @>` **fail**:

The quoted expression in that passing test is `<@ Seq.forall id l @>`, where `l` is a `seq<bool>` (and `id` is the built-in `id` function). This quoted expression *doesn't use FsCheck*, but only types and functions from *FSharp.Core*.

All the other properties using `test <@ ... @>` use types or functions from FsCheck in the quoted expressions. This could explain why all of these properties fail, while that passing property passes.

This seems consistent with the error messages, which are all identical:

> System.IO.FileNotFoundException : Could not load file or assembly 'FsCheck, Version=2.1.1.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.

It seems like `test <@ ... @>` is trying to load the assemblies defining the types and functions within the expression, and for some reason, on the AppVeyor build, it can't find FsCheck (the System Under Test)...

ploeh:
@kurtschelfthout, it looks like the AppVeyor build runs on Windows, whereas the Travis build runs on Linux. Is that a correct interpretation?

ploeh:
Following up on the Windows/Linux hypothesis, I ran the build on my Windows 10 laptop, in three different environments:

- `./build.sh` from Git Bash (MINGW64): **pass**
- `.\build.cmd` from Windows PowerShell: **pass**
- `build.cmd` from an ordinary Command Prompt: **pass**

It'd be interesting to see if anyone else is able to reproduce this build error.

moodmosaic:
I'll take a look on my machine as well, and report back.

kurtschelfthout:
Yes: appveyor is windows and .NET, travis is linnux and mono.

I would actually be ok with lifting out the FsCheck expression out of the `test` expressions. I don't think Unquote can do anything sensible with them anyway.

moodmosaic:
FWIW, on my machine as well, [the build passes](https://twitter.com/nikosbaxevanis/status/651311826932789248) on MINGW64, PowerShell, and the ordinary command-prompt (cmd.exe).

kurtschelfthout:
Must be something on AppVeyor specifically. Strange, don't know what it could be.

moodmosaic:
Those pesky build servers :grin:

kurtschelfthout:
@ploeh I would add a change that sets prints this property to the console on those failures to figure out what is going on on appveyor. We can always undo or remove it later. https://msdn.microsoft.com/en-us/library/system.io.filenotfoundexception.fusionlog(v=vs.110).aspx

kurtschelfthout:
Actually, perhaps it is easier if I merge the change, and then you can open another "throwaway" to figure out what is going on on appveryor. I'm fine with either, what do you think?

stephen-swensen:
Hi guys - I've not seen this particular issue before and am not familiar with AppVeyor. But looking at the stack trace, it appears that the exception originates from the quotation itself and may not necessarily be directly related to Unquote. i.e. note the transition from `Microsoft.FSharp.Quotations` to `System.Reflection.Assembly`

```
...
at System.Reflection.Assembly.Load(String assemblyString)
at Microsoft.FSharp.Quotations.PatternsModule.u_assref(InputState st)
...
```

You could test this hypothesis by writing a test like `ignore <@ fscheck type captured in here @>` and see if it fails similarly without Unqoute's involvement.

The "good" news is that I don't spot any custom `AssemblyResolve` handlers in the stack trace (like [this one](http://stackoverflow.com/questions/2024036/strange-fsi-exe-behavior) I recently ran into), so it appears that we are dealing with "normal" assembly binding failure, and hopefully the `FusionLog` can help pin point the issue - something subtle no doubt!

jhamm:
This can be reproduced by running the tests from FAKE, at least it does from my machine :smile:.

jhamm:
I found that this appears in the Assembly Binding Log Viewer `fuslogvw.exe`.
xunit.console.exe is looking for FsCheck.dll in the same directory as the xunit.console.exe.

```
*** Assembly Binder Log Entry  (10/7/2015 @ 12:59:17 AM) ***

The operation failed.
Bind result: hr = 0x80070002. The system cannot find the file specified.

Assembly manager loaded from:  C:\Windows\Microsoft.NET\Framework64\v4.0.30319\clr.dll
Running under executable  {PATH}\FsCheck\packages\xunit.runner.console\tools\xunit.console.exe
--- A detailed error log follows. 

=== Pre-bind state information ===
LOG: DisplayName = FsCheck, Version=2.1.1.0, Culture=neutral, PublicKeyToken=null
 (Fully-specified)
LOG: Appbase = file:///{PATH}/FsCheck/packages/xunit.runner.console/tools
LOG: Initial PrivatePath = NULL
LOG: Dynamic Base = NULL
LOG: Cache Base = {Removed}
LOG: AppName = 95d0aa47-9a64-43ba-b940-77ccf207ce1b
Calling assembly : (Unknown).
===
LOG: This is an inspection only bind.
LOG: Using application configuration file: {PATH}\FsCheck\packages\xunit.runner.console\tools\xunit.console.exe.Config
LOG: Using host configuration file: 
LOG: Using machine configuration file from C:\Windows\Microsoft.NET\Framework64\v4.0.30319\config\machine.config.
LOG: Attempting download of new URL file:///{PATH}/FsCheck/packages/xunit.runner.console/tools/FsCheck.DLL.
LOG: Attempting download of new URL file:///{PATH}/FsCheck/packages/xunit.runner.console/tools/FsCheck/FsCheck.DLL.
LOG: Attempting download of new URL file:///{PATH}/FsCheck/packages/xunit.runner.console/tools/FsCheck.EXE.
LOG: Attempting download of new URL file:///{PATH}/FsCheck/packages/xunit.runner.console/tools/FsCheck/FsCheck.EXE.
LOG: All probing URLs attempted and failed.
```

kurtschelfthout:
Thanks @jhamm! That gave me the necessary hint, I just pushed: https://github.com/fscheck/FsCheck/commit/d203f9900326e89fe131369e308c57ecdea6bd30

@ploeh: if you merge from master now I believe this PR should work on AppVeyor.

ploeh:
Uh-oh, it's almost like we flipped a switch. Now the AppVeyor build succeed, but the Travis build failed... :disappointed:

ploeh:
@kurtschelfthout, I had a look at d203f99, and there's a couple of things I don't understand:

- Where's the value `isLinux` defined? When I look at the complete file, I can only find it twice, and in none of those places is it defined or passed in as an argument. Perhaps there's something about FAKE that I'm not aware of, but to me, it looks like F# code that oughtn't compile.
- The comment explaining the `NoAppDomain` setting seems to say that the value should always be true. Am I reading the comment wrong?

kurtschelfthout:
@ploeh 
I fixed the comment.

I have every expectation that the travis build will work, currently it's failing because it can't seem to download the mono packages - nothing to do with your PR. I will retry the build tonight before I merge, but I don't think at this point you or anyone should continue to spend time on this.

kurtschelfthout:
So green. Thanks!

ploeh:
Wow, thank you, all, @kurtschelfthout, @jhamm, @stephen-swensen, @moodmosaic! This was a true community effort, clearly demonstrating the power of open source development :tada: 

I don't think anyone of us alone could have solved this issue as quickly as now, when each was able to contribute their piece of the puzzle.

moodmosaic:
:tada:

PR #165, opened 2015-10-08
Format test code according to 'Arrange, Act, and Assert' heuristic

### LINKED ISSUE #163: Make the property-based tests easier to read
Now that we use Unquote (which is great!), we could make the property-based tests easier to read.

As @ploeh wrote in #162:

>With the built-in operators `=!` and `<>!`, the assertions become so terse that they almost 'drown' in the test. If we stick with these operators, I would recommend adopting this heuristic for formatting tests: http://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern

I could happily send a Pull Request during the next couple of days, if you'd like to follow this direction.

moodmosaic:
This Pull Request resolves #163, by adopting [this](http://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern) useful heuristic for formatting tests.

Each commit log's message contains information regarding the motivation of each particular formatting.

### DIFF
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -188,6 +188,7 @@ module Gen =
             |> sample1
             |> List.replicate 10
             |> List.map ((|>) ())
+
         generated
         |> Seq.pairwise
         |> Seq.map (fun (a,b) -> a = b)
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -285,6 +285,7 @@ module Gen =
                     s := !s + 1
                   return !s
                 }
+
         test <@ convolutedGenNumber 100 
                 |> Gen.sample 1 10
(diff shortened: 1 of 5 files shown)


ploeh:
Looks good to me :+1:

kurtschelfthout:
Can you do me a favour and squash this in a single commit?

moodmosaic:
The commit messages may look similar, but they are not identical.

If I squash all in a single commit, the *exact* reasoning behind each change will get lost. For example, the message of e1320db is different than e968858, and so on.

kurtschelfthout:
I'm fine with losing the exact reasoning.

moodmosaic:
Sorry, but I wouldn't like to move forward in the way you propose :disappointed: – In my opinion, the reasoning behind each change matters here, as we're following a heuristic.

You're the maintainer of this codebase, so I won't push further, thus I'll just go ahead and close this...

Thanks for you time!

PR #170, opened 2015-10-18
Migrate to paket

sideeffffect:
WIP

PR #171, opened 2015-10-19
WIP: Use Paket Pack

forki:
This tries to make #170 work

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -198,6 +198,26 @@ Target "SourceLink" (fun _ ->
 // --------------------------------------------------------------------------------------
 // Build a NuGet package
 
+let projectDirs = List.map (sprintf "src/%s")
+                           [ "FsCheck"
+                             "FsCheck.NUnit"
+                             "FsCheck.Xunit"
+                           ]
+
+Target "PaketPack" (fun _ ->
+  projectDirs |> Seq.iter (fun projectDirs ->
+    Paket.Pack (fun p ->
+      { p with
+          TemplateFile = "paket.template"
+          WorkingDir = projectDirs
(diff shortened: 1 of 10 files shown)


build.fsx
```diff
@@ -305,10 +297,12 @@ Target "CI" DoNothing
...
-  ==> "NuGet"
+  ==> "PaketPack" 
+  ==> "PaketPush"
   ==> "Release"
 
-"GenerateDocs" ==> "CI"
-"NuGet" ==> "CI"
+"GenerateDocs"
+  ==> "PaketPack" 
+  ==> "CI"
```
kurtschelfthout: Not a big deal but I like to have separate "generate nuget packages" and "generate documentation" targets, i.e. if I'm checking the packaging I don't necessarily want to generate the docs.

build.fsx
```diff
@@ -198,24 +196,19 @@ Target "SourceLink" (fun _ ->
...
+          OutputPath = "../../bin/PaketPack"
+          Version = buildVersion
+          ReleaseNotes = toLines release.Notes
+      })
+)
+
+Target "PaketPush" (fun _ ->
+    Paket.Push (fun p -> 
+        { p with 
+            WorkingDir = "../../bin/PaketPack" })
```
kurtschelfthout: Does this work the same way with regards to the nugetkey? I.e. clearly the push won't work unless it knows my key, but do I still set is as an env variable in my (not checked in) release script?

kurtschelfthout:
This is looking really good. Thanks a bunch!

build.fsx
forki: Yes it uses same conventions as before

PR #172, opened 2015-10-19
Finish migrating to Paket

sideeffffect:


sideeffffect:
```
Paket failed with:
	The file "bin/Release/FsCheck.pdb" requested in /home/travis/build/fscheck/FsCheck/src/FsCheck/paket.template does not exist.
```

it seems like the `*.pdb` and `*.xml` files are available on *.NET*, but are not on *Mono*
my question is, do we need them in the resulting `nupkg`?

kurtschelfthout:
Yes, we need them. Otherwise people won't be able to see the docs, and the pdb files are needed to make source linking work.

sideeffffect:
ok, how can we create these files in Mono? they don't get created on my system or travis :(

kurtschelfthout:
Don't know, seems like mono should be able to create these things.

build.fsx
```diff
@@ -288,21 +236,24 @@ Target "CI" DoNothing
...
 
 "GenerateDocs"
+  ==> "CI"
+
+"RunTests"
   ==> "PaketPack" 
   ==> "CI"
 
-RunTargetOrDefault "RunTests"
+RunTargetOrDefault "CI"
```
kurtschelfthout: Please leave the default target as RunTests.

PR #176, opened 2015-11-03
Update xUnit summary on property failure due to exception

neoeinstein:
This initially comes from the xUnit runner failing to increment the failure count on an exception being thrown during property evaluation (as might happen when using Unquote). In this case, the test was not being added to the summary as failed, though that was the test result returned. The number of failures reported ends up being xUnit's exit code, so if a property ended up throwing an exception and failing, it might not be seen as a failure to an external program (such as FAKE) that relied on that exit code.

This change increments those failure counters on a thrown exception so that it appropriately influences xUnit's exit code and matches user expectations in the console. This PR is also a bit of a question: it looks like these two cases were singled out for not incrementing the failure count, so it may be for a reason that I don't know about.


kurtschelfthout:
Good point I think, it looks like an oversight to me.

@ploeh any thoughts?

ploeh:
Looks legit to me, but I can't say that I understand the xUnit.net 2 way of doing things...

PR #180, opened 2015-11-09
Performance optimizations

mrange:
Performance optimization for F# Unions
==============================

When using FsCheck to generate test cases for my optimized Tries I thought FsCheck was taking too long. 

After some profiling I noticed FsCheck spent most of its time in: `getUnionCases`. In addition I noticed non-working memoization in `reflectObj`

So I:

1. Added memoization to `getUnionCases`
1. Fixed broken memoization of `reflectObj`
1. Switched to `ConcurrentDictionary<>` for `memoize`
   Doesn't lock on lookups
   Portable binaries still uses lock

Performance numbers previous to change:
6x100 property tests  ==> ~30 sec

Performance numbers after change:
6x100 property tests  ==> ~1 sec

Your milage might differ depending on how much your test cases calls `getUnionCases`/`reflectObj`

Regard,
Mårten

### DIFF
--- a/src/FsCheck/Common.fs
+++ b/src/FsCheck/Common.fs
@@ -12,7 +12,12 @@ namespace FsCheck
 
 module internal Common =
 
+    open System
+#if PCL
     open System.Collections.Generic
+
+    ///Create the Memoizer
+    let memoizer () = new Dictionary<_,_>()
 
     ///Memoize the given function using the given dictionary
     let memoizeWith (memo:IDictionary<'a,'b>) (f: 'a -> 'b) =
@@ -25,12 +30,22 @@ module internal Common =
                         let res = f n
                         memo.Add(n,res)
                         res)
+#else
+    open System.Collections.Concurrent
+
+    ///Create the Memoizer
+    let memoizer () = new ConcurrentDictionary<_,_>()
+
+    ///Memoize the given function using the given dictionary
+    let memoizeWith (memo:ConcurrentDictionary<'a,'b>) (f: 'a -> 'b) =
+        let ff = Func<'a,'b>(f)
+        fun n ->
+            memo.GetOrAdd (n, ff)
+#endif
 
     ///Memoize the given function.
     let memoize f =
-        let t = new Dictionary<_,_>()
-        memoizeWith t f
-
+        memoizeWith (memoizer ()) f
     let flip f x y = f y x
 
     //the following three are from Don Syme's blog:
--- a/src/FsCheck/Reflect.fs
+++ b/src/FsCheck/Reflect.fs
@@ -60,13 +60,15 @@ module internal Reflect =
         ctor.Invoke
 
     /// Returns the case name, type, and functions that will construct a constructor and a reader of a union type respectively
-    let getUnionCases unionType : (string * (int * System.Type list * (obj[] -> obj) * (obj -> obj[]))) list = 
+    let getUnionCases : Type -> (string * (int * System.Type list * (obj[] -> obj) * (obj -> obj[]))) list =
+        Common.memoize (fun (unionType:Type) ->
         [ for case in FSharpType.GetUnionCases(unionType, true) ->
             let types =    [ for fld in case.GetFields() -> fld.PropertyType ]
             let ctorFn =   FSharpValue.PreComputeUnionConstructor(case, true)
             let readerFn = FSharpValue.PreComputeUnionReader(case, true)
                 
             case.Name, (case.Tag, types, ctorFn, readerFn)]
+          )
 
     /// Get reader for union case name (aka tag)
     let getUnionTagReader unionType = 
--- a/src/FsCheck/ReflectArbitrary.fs
+++ b/src/FsCheck/ReflectArbitrary.fs
@@ -33,22 +33,24 @@ module internal ReflectArbitrary =
        else 
            elems
 
-    let private reflectObj getGenerator =
+    let private reflectObjGenerator =
         Common.memoize (fun (t:Type) ->
             if isRecordType t then
-                let g = [ for pi in getRecordFields t do 
-                            if pi.PropertyType = t then 
-                                failwithf "Recursive record types cannot be generated automatically: %A" t 
-                            else yield getGenerator pi.PropertyType ]
+                let fts = [|for pi in getRecordFields t do
+                                if pi.PropertyType = t then
+                                    failwithf "Recursive record types cannot be generated automatically: %A" t
+                                    else yield pi.PropertyType|]
                 let create = getRecordConstructor t
-                let result = g |> sequence |> map (List.toArray >> create)
-                box result
+                fun getGenerator ->
+                    let result = fts |> Seq.map getGenerator |> sequence |> map (List.toArray >> create)
+                    box result
 
             elif isTupleType t then
-                let g = [ for elem in FSharpType.GetTupleElements t do yield getGenerator elem ]
+                let fts = [| for elem in FSharpType.GetTupleElements t do yield elem |]
                 let create elems = FSharpValue.MakeTuple (elems,t)
-                let result = g |> sequence |> map (List.toArray >> create)
-                box result
+                fun getGenerator ->
+                    let result = fts |> Seq.map getGenerator |> sequence |> map (List.toArray >> create)
+                    box result
 
             elif isUnionType t then
                 // figure out the "size" of a union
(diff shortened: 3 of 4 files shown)


src/FsCheck/Reflect.fs
```diff
@@ -60,13 +60,15 @@ module internal Reflect =
         ctor.Invoke
 
     /// Returns the case name, type, and functions that will construct a constructor and a reader of a union type respectively
-    let getUnionCases unionType : (string * (int * System.Type list * (obj[] -> obj) * (obj -> obj[]))) list = 
+    let getUnionCases : Type -> (string * (int * System.Type list * (obj[] -> obj) * (obj -> obj[]))) list =
```
mrange: Possible a breaking ABI change. I will inspect the generated binaries to if it changes the signature. If so changing the signature back to previous one isn't a big deal.

src/FsCheck/Common.fs
```diff
@@ -25,12 +30,22 @@ module internal Common =
...
                         memo.Add(n,res)
                         res)
+#else
+    open System.Collections.Concurrent
+
+    ///Create the Memoizer
+    let memoizer () = new ConcurrentDictionary<_,_>()
+
+    ///Memoize the given function using the given dictionary
+    let memoizeWith (memo:ConcurrentDictionary<'a,'b>) (f: 'a -> 'b) =
```
mrange: ConcurrentDictionary only uses locks on inserts IIRC. A minor improvement for FsCheck as there should be little contention in test code. Still nice IMO

src/FsCheck/ReflectArbitrary.fs
```diff
@@ -33,22 +33,24 @@ module internal ReflectArbitrary =
        else 
            elems
 
-    let private reflectObj getGenerator =
```
mrange: Previous reflectObj memoization rarely seemed to succeed. Rewrote in order to cache most of the operation and apply getGenerator on the cached info.

src/FsCheck/ReflectArbitrary.fs
```diff
@@ -59,40 +61,49 @@ module internal ReflectArbitrary =
...
-                let gs = [ for _,(_,fields,create,_) in getUnionCases t -> unionSize fields, lazy (unionGen create fields) ]
-                let lowest = List.reduce min <| List.map fst gs
-                let small() = [ for i,g in gs do if i = lowest then yield g.Force() ]
-                let large() = [ for _,g in gs -> g.Force() ]
-                let getgs size = 
-                    if size <= 0 then small() else large() 
-                    |> oneof 
-                    |> resize (size - 1) 
-                sized getgs |> box
+                let gs = [| for _,(_,fields,create,_) in getUnionCases t -> unionSize fields, unionGen create fields |]
```
mrange: Most changes were trivial but this was a bit tricky. So here I really like some feedback.

kurtschelfthout:
Cool! Let me look through this, might take a bit of time...

mrange:
Np, I am not in a rush. The change to `getUnionCases` is what most important in my case if you get unsure about the `reflectObj` change (which was trickier)

kurtschelfthout:
Yes, I think it would speed up the process and allow easier discussion if you split this up in 3 PRs, or at least one PR but with 3 separate commits. Then I can easily test each separately too.

What do you think?

mrange:
1 commit => 3 commits, done

kurtschelfthout:
Can you give me one or more code examples where this helps?  I'm having a bit of trouble reproducing your results.

For example, the following runs in the same time for me before and after your change:

    type UnionCase =
    | Single
    | One of int
    | Two of string * int
    | Three of string * char * int
    | Rec of int * UnionCase

    let run() =
        let res = Arb.from<UnionCase>.Generator |> Gen.sample 100 100000
        sprintf "%i" res.Length

mrange:
I have a branch with the actual code I ran into performance problems:

https://github.com/mrange/FsCheck/tree/mrange/perf_test

On my machine on release builds I see a significant improvement.

Let me see if I can understand why your example can't reproduce the improvement.

mrange:
The project is called FsCheck.PerfTest

mrange:
`Arb.from<UnionCase>.Generator |> Gen.sample 100 100000` seems only to invoke `getUnionCases` 5 times for the same type. For the code I posted `getUnionCases` is invoked many times (1000+) for the same type.

It's possible that I am doing something wrong in my testcase. I am not even an FsCheck beginner.

mrange:
Btw with a slight rewrite of `frequency` in Gen.fs it seems I got a 15% improvement for your sample.

Not sure how it maps to realistic settings:
```
        // Arrays are generally faster than Seq.cache
        let xs = Seq.toArray xs
        let tot = Array.sumBy fst xs  // And sumBy is evaluating the how sequence anyway
        let rec pick i n =
            let k,x = xs.[i]
            if n<=k then x else pick (i+1) (n-k)
        gen.Bind(choose (1,tot), pick 0)
```

kurtschelfthout:
OK, I'll take a look at that.

mrange:
Did a smaller repro. This code performs ~5x better for me on the PR

```
type UnionCase =
  | Single
  | One of int
  | Two of string * int
  | Three of string * char * int
  | Rec of int * UnionCase
  member x.CheckInvariant () =
    match x with
    | Single          -> true
    | One _           -> true
    | Two (s,_)       -> s <> null
    | Three (s, _, _) -> s <> null
    | Rec (_,_)       -> true

let testProp (uc : UnionCase) =
  true

let testPerf () =
  let config =
    {
      Config.Quick with
        MaxTest = 1000
    }
  Check.One (config, testProp)
```

kurtschelfthout:
Yes, huge difference, just with the getUnionCases memo. That suggests something more insidious is going wrong though - it means it is re-creating the generator too many times, or something. I need to take a deeper look.

kurtschelfthout:
Got one: https://github.com/fscheck/FsCheck/commit/b0689d6bf0951288441c1c08f19db1280db3bc50

mrange:
I will try the latest master

kurtschelfthout:
With the change above in, the commits proposed here don't seem to make any difference, perhaps even making things slight worse. Do you see the different?

Thanks for the hint on the pick - that sounds like a good change. Do you want to PR that one separately for the credit?

mrange:
Tried latest master. Gave ~6x improvement in my use case. For my use case the complete PR gives around 20x-30x performance but it might be an odd use case.

mrange:
Let me investigate what gives the difference for me and we can put the PR on hold.

kurtschelfthout:
Ok, cool. I think it may be related to the fact that you have properties with multiple parameters, which causes effectively nested `forAll` calls, which will mean my optimization only takes effect on the outer one.

I am happy to look at your full use case tomorrow, doesn't look too complicated. In case you have better things to do :)

Getting some sleep now...

kurtschelfthout:
This is cool though, wanted to look at perf for a while now. Not surprised there is much low hanging fruit here! Definitely appreciate the effort.

mrange:
Uploaded the `frequency` improvement as a separate PR.

kurtschelfthout:
Made some progress on this yesterday - another 5x improvement by avoiding to recreate the reflective generator each time. It caused a few test breaks though, need to investigate how to handle.

kurtschelfthout:
I think I finally got it. You were on the right track with the broken memoization, but the fix above didn't really fix it. My commit in https://github.com/fscheck/FsCheck/commit/62bb037c16ccc01829d5b2d9ac25b06481e14dbf does (knocks on wood).

Can you recheck with master now and keep only the commits on this PR that give some improvement? 

* The broken memoization in `reflectObj` can definitely go, I think. 
* The extra memoization in `getUnionCases` may no longer be worth it (but hey, happy to be proven wrong). 
* I am overall a bit skeptical about `ConcurrentDictionary` - while you are technically right I feel this only makes sense under high concurrent load, which this isn't. So the extra complexity is probably not worth it.

mrange:
I have tested against latest master and applied the individual commits of this performance optimization.

Now I can't reliable detect any performance improvement of either commit so I think the value of this PR is non-existant anymore.

Possibly with the exception of `ConcurrentDictionary`. Locks are a pain IMHO (spend my workdays doing realtime programming) and when there's a good replacement (as IMHO `ConcurrentDictionary` is) my reaction is to go for it. Anyway, it's not a big deal in this case probably.

I close this PR now. 

Btw. I estimate ~60x performance improvements for my use case. So that's pretty decent.

kurtschelfthout:
Thanks for checking, that's great. This should have similar effects of reflective generation of records etc. 

I don't disagree on the concurrent dictionary. Use it all the time in the day job too. In this case however the concurrent calls are very light (just running xunit tests in parallel) and we have the added complexity of the conditional compilation for PCL which I'm trying to minimise. 

Certainly worth keeping an eye on in case the conditions change though. 

Thanks again for raising!

PR #181, opened 2015-11-11
Improved perf of frequency

mrange:
Rewrote `frequency` to a more efficient pattern. As `frequency` seems heavily used this improved performance in some runs for me as much as 15%.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -180,13 +180,13 @@ module Gen =
     ///given probabilities. The sum of the probabilities must be larger than zero.
     //[category: Creating generators from generators]
     [<CompiledName("Frequency")>]
-    let frequency xs = 
-        let xs = Seq.cache xs
-        let tot = Seq.sumBy fst xs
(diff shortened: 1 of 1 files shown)


src/FsCheck/Gen.fs
```diff
@@ -180,13 +180,13 @@ module Gen =
...
-    let frequency xs = 
-        let xs = Seq.cache xs
-        let tot = Seq.sumBy fst xs
-        let rec pick n ys = 
-            let (k,x),xs = Seq.head ys,Seq.skip 1 ys
-            if n<=k then x else pick (n-k) xs
-        in gen.Bind(choose (1,tot), fun n -> pick n xs) 
+    let frequency xs =
+        let xs = Seq.toArray xs
+        let tot = Array.sumBy fst xs
```
mrange: As sumBy is evaluating the full sequence any it is often more efficient to store the result as an array. Then pick iterates over this array using array indices.

kurtschelfthout:
Thanks!

PR #183, opened 2015-11-12
Improved perf of getCSharpRecordConstructor

mrange:
### Performance improvment for getCSharpRecordConstructor ###

Used `System.Linq.Expressions` to compile delegates that are more efficient
than `ConstructorInfo.Invoke`.

This is added complexity to the code which might be discouraging. I can sympatize with a reluctance to increase complexity. I thought I bring up this as a possibility to improve reflection performance. There are other places in FsCheck that could benefit from this technique if it is deemed interesting.

Anyway, during my test runs which relied on generating lots of instances this improved performance ~30%. For realistic scenarios this probably lower as the overhead from `.ctor` invoke is less.

My test code:
```
type MyClass(n: int) =
  let x = n + 1

let testCtorPerf () =
  let g = Arb.from<MyClass>.Generator

  let ms, _ = timeIt (fun () ->
    for i in 0..2000 do
      g |> Gen.sample 100 2000 |> ignore
    )

  hilightf "Checking CtorPerf took: %d ms" ms
```

### DIFF
--- a/src/FsCheck/Reflect.fs
+++ b/src/FsCheck/Reflect.fs
@@ -16,6 +16,7 @@ module internal Reflect =
     open System.Collections.Generic
     open Microsoft.FSharp.Reflection
     open System.Reflection
+    open System.Linq.Expressions
 
 //    let private recordFieldBindingFlags = 
 //        BindingFlags.GetProperty ||| BindingFlags.Instance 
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
I think that's cool for now. I'm actually surprised it works in PCL. Let me double check the results.

Longer term I was thinking of replacing the reflection calls with the ones in FSharpx.Reflection, which should be much faster.

https://github.com/fsprojects/FSharpx.Reflection

mrange:
`MethodInfo.Invoke` is "fixable" in the same way

PR #184, opened 2015-11-15
Generation of docs is possible on Mono

sideeffffect:


kurtschelfthout:
Well, no, it's not. Granted, the build doesn't fail, but there are multiple exceptions and errors. It can't create some directory, there is no FsCheck.xml file,...

If you want to work on that you can use this PR to test, but I'm not going to merge the change until it actually works :)

PR #186, opened 2015-11-18
Allow reflector to derive generator for unions with a private constructor

neoeinstein:
We can do it for records, so let's also do it with unions. Tests included for the arbitrary generator for both private unions and private records.


kurtschelfthout:
Yes, let's. Thanks!

PR #194, opened 2015-12-07
support shrinking null values for C# compatibility

willryan:
I ran into an issue writing tests in C# for data transfer objects defined in F#.  I've found that F# records make good DTOs because of deep equality and immutability.  However, if you are referring to F# records in C#, records and object types can have null values even when this is not allowed in F# itself.  Since null is an acceptable value in C#, you should be able to generate nulls and return them as an option for a shrunk value.  This would all be possible if I wrote a complete shrinker, but I'd much rather leverage the existing shrinker for F# records when possible.  However if you try to shrink a null value for a field in an F# record, FsCheck crashes because you cannot unbox a null into an F# type unless the AllowNullLiteral attribute is set.

You could argue that if you intend to expose your F# records in C# you should always specify the AllowNullLiteral attribute.  However, since this attribute is set on the type and not on its references, you cannot control under which circumstances you wish to allow null.  In any event, it is fairly trivial to check for null and infer that additional shrinking is impossible.  Or in the more general case, if a value cannot be unboxed (essentially implying an object state that F# does not consider valid), additional shrinking is impossible.  That's the solution I've opted for here.

Let me know if you'd prefer another way of dealing with this issue, or if there are any changes you'd like me to make.

### DIFF
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -7,6 +7,7 @@ module Gen =
     open FsCheck.Xunit
     open Helpers
     open Swensen.Unquote
+    open System.Reflection
 
     [<Property>]
     let Choose (Interval (l,h)) = 
@@ -180,6 +181,29 @@ module Gen =
             (Array2D.length1 arr <= rows) 
             && (Array2D.length2 arr <= cols) 
             && (seq { for elem in arr do yield elem :?> int} |> Seq.forall ((=) v))
+
+
+    type MaybeNull =
+        {
+            SomeField : int
+        }
+
+    [<CLIMutable>]
+    type MaybeHaveNull =
+        {
+             MightBeNull : MaybeNull
+        }
+
+    [<Fact>]
+    let ``should support null as a shrunken value even without AllowNullLiteral (for C# compatibility)``() =
+        let der = Arb.Default.Derive()
+        let mhn = { MightBeNull = { SomeField = 5 } }
+        let prop = mhn.GetType().GetProperty("MightBeNull", BindingFlags.Public ||| BindingFlags.Instance)
+        prop.SetValue(mhn, null)
+        // ensure that no exception occurs
+        der.Shrinker mhn
+          |> Seq.length
+          |> ignore
 
     [<Fact>]
     let ``should generate functions from nullable values``() =
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -39,7 +39,18 @@ type Arbitrary<'a>() =
         Seq.empty
     interface IArbitrary with
         member x.GeneratorObj = (x.Generator :> IGen).AsGenObject
-        member x.ShrinkerObj o = (x.Shrinker (unbox o)) |> Seq.map box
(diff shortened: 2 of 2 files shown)


tests/FsCheck.Test/Gen.fs
```diff
@@ -181,6 +182,29 @@ module Gen =
...
+    [<Fact>]
+    let ``should support null as a shrunken value even without AllowNullLiteral (for C# compatibility)``() =
+        let der = Arb.Default.Derive()
+        let mhn = { MightBeNull = { SomeField = 5 } }
+        let prop = mhn.GetType().GetProperty("MightBeNull", BindingFlags.Public ||| BindingFlags.Instance)
+        prop.SetValue(mhn, null)
+        // ensure that no exception occurs
+        der.Shrinker mhn
+          |> Seq.length
+          |> ignore
```
kurtschelfthout: I may be missing something, but this seems like a very roundabout way of testing the change. How about something like:

     let shrinker = Arb.Default.Derive().Shrinker
     shrinker null |> Seq.length != 0

kurtschelfthout:
Thanks, it's good to make the shrinker more robust. 

Raises a few other issues (not that I'm asking you to fix those!) like we should probably generate null values for F# types that have an `AllowNullLiteral` attribute.

tests/FsCheck.Test/Gen.fs
willryan: One might be able to make this test a bit simpler, but unfortunately the example you provided won't quite do it.  The null value is inferred to have the 'object' type, so reflectShrinkObj does not recognize a record and therefore does not end up calling shrinkChildren.  shrinkChildren is what calls getShrink, which is what calls to ShrinkerObj (where I made the code change).  
Moreover, if you send in a record with an (illegitimate) null value directly to the derive shrinker, it will crash in a different way than I was seeing (the outer loop in shrinkChildren's seq generator).  I'm not sure whether that case should be considered a problem, since one generally does not invoke the shrinker directly, and sending it a null value that is recognized as something other than 'object' requires some deliberate misbehavior.  My test case attempted to replicate my problem as closely as possible, but in F# instead of the C# where I was seeing it.  But let me know if you still want me to try to find a simpler test case.

willryan:
Also, for what it is worth I created some extension functions for dealing with the null generation and shrinking in C#.  What are your thoughts on a FsCheck C# Extensions nuget project?  It would have C#-specific functionality, and has regular FsCheck as a dependency?  I only have a small use case for it now, but it might grow in the future.  I might have a use case coming up which would involve making a generator and shrinker for objects tagged with DataContract.

kurtschelfthout:
You should do whatever you feel best about!

That said, since you asked, personally I encourage you to contribute directly to FsCheck itself. I am ok with adding "C# specific" features to FsCheck directly. 

I think it's more user-friendly for users if they get what they need in one library, rather than have to look around for something. It also makes distribution more efficient, and there is less opportunity for error (e.g. version mismatches).

tests/FsCheck.Test/Gen.fs
kurtschelfthout: Ok, not a big deal.

Issue #203, opened 2016-03-13
Config and Configuration


MarcoMaar:
I have a question regarding the configurations that are used to Check a property

For example at ```Check.One<Testable>(Config config, Testable property)``` a configuration of the type ```Config``` is used where all properties of are readonly. 

But at ```CheckExtensions.CheckCheck(this Property property, Configuration config)``` a configuration of the type ```Configuration``` is used where you can change the values, however the ```CheckExtensions.VerboseCheck``` extension does have an overload that takes a ```Configuration```instance

I am curious why ```Config``` is read only, why there are two types of configurations and why ```CheckExtensions.VerboseCheck``` does not take a configuration .

For me it would be convenient if there is only one configuration , where it is possible to change single properties of it.

mausch:
The mutable `Configuration` and associated extension methods are meant for C# . Once you pass your `Configuration` instance to e.g. `CheckExtensions.Check` it is converted to the immutable `Config` type.
If you're writing your tests in F#, use the immutable `Config` type. F#'s copy-and-update syntax makes it trivial to change single properties (giving you a new instance of course).

MarcoMaar:
Okay, that makes sense. Thank you.
I am still wondering why the Verbose functions don't have an overload with a configuration.

kurtschelfthout:
Because verbosity configures, via `Config` or via `Configuration` what gets printed when certain events happen (e.g. value tested, value shrunk, etc). And so Quick and Verbose are just two particular configurations. So basically a method Check.Verbose or Check.Quick that takes a configuration is ambiguous (because it would somehow need to "merge" two configurations, the verbose one and the given one).

I guess we could add some static methods to `Configuration` that return the quick, verbose and default `Configuration` instances; we have a similar thing on `Config`.

MarcoMaar:
I just updated my previous answer the moment you wrote.
If it is okay to add those static methods that would be nice. Will you take care of it or should I do that?

kurtschelfthout:
Sure, go for it.

kurtschelfthout:
Released in 2.2.5.

PR #204, opened 2016-03-15
Static Configuration

MarcoMaar:
PR for Issue #203 
Added static configuration instances (default, quick, verbose + throw…… on failure). 
Wrote two test cases in CounterTest to see the behaviour

### DIFF
--- a/src/FsCheck/RunnerExtensions.fs
+++ b/src/FsCheck/RunnerExtensions.fs
@@ -13,7 +13,7 @@ namespace FsCheck
 open System
 open System.Runtime.CompilerServices
 
-///Configure the test run.
+///Configure the test run with a quick configuration.
 type Configuration() =
     let mutable maxTest = Config.Quick.MaxTest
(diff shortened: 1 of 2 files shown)


src/FsCheck/RunnerExtensions.fs
```diff
@@ -26,6 +26,33 @@ type Configuration() =
     let mutable runner = Config.Quick.Runner
     let mutable replay = Config.Quick.Replay
 
+    ///val get initialitzed with Config.Quick anyway
```
kurtschelfthout: This is a code comment, should not be xml doc.

kurtschelfthout:
Thanks for the PR!

MarcoMaar:
okay, changed the XML doc.
Not sure though why the build fails.

kurtschelfthout:
Hm, yes, weird error. I think you are just unlucky :) I will investigate the failing test separately.

Issue #208, opened 2016-03-22
Experimental.StateMachine Integrating externally generated sequences


MarcoMaar:
I am trying to combine a model based testing strategy with FsCheck because I suspect it to generate better ```Command``` Sequences. My other tool would generate the type of ```Command``` (lets call it ```TypeSequence```) while FsCheck would generate the members of it.

I am not sure how to implement this with the current version of FsCheck. I could leverage the ```Next``` function to always pick the next element from my ```TypeSequence```. But I don't know how to make sure that it takes all elements from it, since I cannot control the exact length used for generating the ```CommandSequence``` in FsCheck.

I understand that my intend arguably goes against the purpose of FsCheck. I am using a "random generator" but trying to leverage some of the randomness. But I think it would be a great addition if we can somewhat control the ```CommandSequence``` used in the ```StateMachine```. It allows us to combine it with other TestCase Generators.

So my question are: Am I missing something and it is already possible to implement what I want? If no, would it be an option to support using external Generators for ```Commands```?

kurtschelfthout:
I think if you want FsCheck to generate something to use with another tool, you can write the generator then use `Gen.sample` to get a bunch of elements.

There is no way to tell FsCheck to generate elements from a list exhaustively - the assumption is that exhaustive testing is not possible because the state space is too large. (Not saying this isn't useful)

MarcoMaar:
Maybe I explained poorly again.  I do not want to do exhaustive testing, I want to tell FsCheck which Commands to use in Next.

I still would like to use the  ```Machine<'Actual, 'Model>```.
With ```CommandSequence``` I mean the commands that will be generated with the ```Next```. They are generated one after the other. Therefore I don't know how to control the size of said Sequence.

I will go with an example.
My external tools generates me a sequence of Transitions. e.g..  Add,Substract,Add,Add each of those transitions will need a parameter(int). I want to use this generated Sequence in a way that the first ```Next``` call in ```Machine``` takes the first Element in my list(Add) and then generates the parameters and then creates a ```Generator``` of  ```AddCommand``` and returns it in ```Next```. Second call uses second element, etc. until all elements are used.

I am very focused on Next here, maybe there is another way to realize what I want, just seemed the Next is currently the closest.

kurtschelfthout:
[...]

PR #210, opened 2016-03-29
PR for Issue #208 StopCommand for StateMachine

MarcoMaar:
I did not find sufficient information on how a stop is implemented into the StateMachine of Erlang's QuickCheck[1,2]. 

I wanted to keep it possible to operate without a StopCommand. For that I created a Property in StateMachine were you can specify the number of Commands for each run. If nothing is specified(value < 0) the size of Gen.sized is used to determine the number of Commands.
Maybe it makes more sense to change it to an overridable function (more flexibility)?

StopCommand is derived from Command and signals the end of the CommandSequence. It is not added to the Sequence though since it is not a real command. Setting a precondition for the StopCommand is possible.

I am not very confident with my f# code since I am not used to write in it. Any suggestions are appreciated.

[1] Hughes, John. "QuickCheck testing for fun and profit." Practical Aspects of Declarative Languages. Springer Berlin Heidelberg, 2007. 1-32.
[2]Csaba, Hoch. "QuickCheck használata szoftverteszteléshez."

### DIFF
--- a/src/FsCheck/StateMachine.fs
+++ b/src/FsCheck/StateMachine.fs
@@ -44,6 +44,12 @@ type Operation<'Actual,'Model>() =
     ///The default precondition is true.
     default __.Pre _ = true
 
+type StopOperation<'Actual,'Model>() =
+    inherit Operation<'Actual,'Model>()
+    override __.Check (a, m) = true.ToProperty()
+    override __.Run m = m
+
+
 type TearDown<'Actual>() =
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
Thanks, I'll comment more on the issue.

For next PR: can you rebase just your change on top of master, instead of merging master to your branch? This keeps the history clean. Most maintainers I think will ask for this.

Issue #220, opened 2016-04-17
Add Gen.shuffle


kurtschelfthout:
Which takes a sequence and generates shuffles of that sequence.

moodmosaic:
May I give this a shot? 😀

kurtschelfthout:
Sure, it's up for grabs. You just grabbed :)

moodmosaic:
Now that #221 is merged, I guess we can close this?

PR #221, opened 2016-04-18
Add Gen.shuffle

moodmosaic:
This Pull Request adds `shuffle` to the `Gen` module, as discussed in #220.

`shuffle` generates a random permutation of a given sequence, and has the type of

```
seq<'a> -> Gen<seq<'a>>
```

It has been ported from Haskell, where `shuffle` appeared in QuickCheck version 2.8.0 ([source](https://hackage.haskell.org/package/QuickCheck-2.8/docs/src/Test-QuickCheck-Gen.html#shuffle)) and got further optimized on version 2.8.2 ([source](https://github.com/nick8325/quickcheck/commit/6c360c1b594b0243e51bfcd4e2428882570e7a51)).

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -313,6 +313,17 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    let shuffle xs =
+        gen {
+            let h = 922337203
+            let gn = choose (-h, h)
+            let! ns = listOfLength (Seq.length xs) gn
+            return xs
+                   |> Seq.zip ns
+                   |> Seq.sortBy fst
+                   |> Seq.map snd
+        }
+
     ///Tries to generate a value that satisfies a predicate. This function 'gives up' by generating None
     ///if the given original generator did not generate any values that satisfied the predicate, after trying to
     ///get values from by increasing its size.
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -313,6 +313,8 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    ///Generates a random permutation of the given sequence.
+    //[category: Creating generators]
     [<CompiledName("Shuffle")>]
     let shuffle xs =
         gen {
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -313,18 +313,22 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
-    ///Generates a random permutation of the given sequence.
+    ///Generates a random permutation of the given sequence using the Fisher-Yates shuffle algorithm.
     //[category: Creating generators]
     [<CompiledName("Shuffle")>]
     let shuffle xs =
+        let xs = xs |> Seq.toArray
+        let swap (arr : _ array) i j =
+            let v = arr.[j]
+            arr.[j] <- arr.[i]
+            arr.[i] <- v
         gen {
-            let h = 922337203
-            let gn = choose (-h, h)
-            let! ns = listOfLength (Seq.length xs) gn
-            return xs
-                   |> Seq.zip ns
-                   |> Seq.sortBy fst
-                   |> Seq.map snd
+            let copy = Array.copy xs
+            let maxI = copy.Length - 1
+            let! indexes = [ for i in 0..maxI - 1 -> choose (i, maxI) ]
+                           |> sequence
+            indexes |> List.iteri (swap copy)
+            return copy
         }
 
     ///Tries to generate a value that satisfies a predicate. This function 'gives up' by generating None
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -313,6 +313,7 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    [<CompiledName("Shuffle")>]
     let shuffle xs =
         gen {
             let h = 922337203
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -120,11 +120,21 @@ module Gen =
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
 
+    // This property is non-deterministic, and may rarely fail. The chance of
+    // this property not holding in case of a uniform shuffle - in other words,
+    // the chance of producing 10 times the same input list of length 5 - is
+    // (1/5!)^10, which should be small enough.
+    // In this case, relying on the statistical attribute of a test being
+    // exceedingly unlikely to produce a False Positive is sometimes the only
+    // option. See also: https://github.com/fscheck/FsCheck/pull/221/
     [<Property>]
-    let Shuffle (NonEmptySet (xs:Set<int>)) =
-        Gen.shuffle xs
+    let Shuffle (s:Set<int>) =
+        s.Count > 1 ==> lazy
+        let l = Set.toList s
+        Gen.shuffle l
         |> sample 10
-        |> List.exists ((<>) (Set.toSeq xs))
+        |> List.map Seq.toList
+        |> List.exists ((<>) l)
    
     [<Property>]
     let SuchThatOption (v:int) (predicate:int -> bool) =
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -119,6 +119,12 @@ module Gen =
         Gen.listOfLength length (Gen.constant v)
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
+
+    [<Property>]
+    let Shuffle (NonEmptySet (xs:Set<int>)) =
+        Gen.shuffle xs
+        |> sample 10
+        |> List.forall ((<>) (Set.toSeq xs))
    
     [<Property>]
     let SuchThatOption (v:int) (predicate:int -> bool) =
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -124,7 +124,7 @@ module Gen =
     let Shuffle (NonEmptySet (xs:Set<int>)) =
         Gen.shuffle xs
         |> sample 10
-        |> List.forall ((<>) (Set.toSeq xs))
+        |> List.exists ((<>) (Set.toSeq xs))
    
     [<Property>]
     let SuchThatOption (v:int) (predicate:int -> bool) =


kurtschelfthout:
Interesting choice, I was expecting a swap algorithm, something like https://github.com/mavnn/MAG/blob/master/MAG/Game.fs#L63

I don't mean it's bad, just surprising :)

I'm a bit worried about using seq. For example, wouldn't the Seq be enumerated at least twice, one for the `Seq.length` and once for the rest? I think an implementation based on `Seq.toArray` in the beginning and then mutating that locally copied array would probably be more efficient. The same pattern is used in `Seq.windowed` for example.

tests/FsCheck.Test/Gen.fs
```diff
@@ -119,6 +119,13 @@ module Gen =
         Gen.listOfLength length (Gen.constant v)
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
+
+    [<Property>]
+    let Shuffle (xs:Set<int>) =
+        not xs.IsEmpty ==> lazy
+        Gen.shuffle xs
+        |> sample 10
+        |> List.exists ((<>) (Set.toSeq xs))
```
kurtschelfthout: This test can be strengthened I think: taking an input list (see `NonEmptyList` and `NonEmptySet`, by the way, to get rid of the precondition) and checking that all the elements in the shuffle appear in the original and vice versa?

src/FsCheck/Gen.fs
```diff
@@ -313,6 +313,19 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    ///Generates a random permutation of the given sequence.
+    //[category: Creating generators]
+    [<CompiledName("Shuffle")>]
+    let shuffle xs =
+        gen {
+            let g = choose (-922337203, 922337203)
```
ploeh: We mere mortals wonder: what's so special about these two magic numbers?

ploeh:
@kurtschelfthout, it may only be a case of parallel evolution, but that implementation looks exactly like something I could have cooked up as well. I have no idea if it's the most efficient implementation, however. I don't hold a CS degree; I'm only a tinkerer.

Regarding concerns about performance, may I suggest, as always, that decisions are made based on measurements, instead of assumptions?

If someone writes up a simple performance comparison script, I'll be happy to run it on my machine(s) in order to report the results.

tests/FsCheck.Test/Gen.fs
ploeh: The only property of `shuffle` that I can think of:

- `List.sort xs` should be equal to `List.sort shuffledXs` (I'm suggesting `List` here, because lists have structural equality, so they are easy to compare.)

We'd also _like_ to be able to state that `xs` should be different from the shuffled sequence, but although random, this property can't be guaranteed - particularly not for small lists.

moodmosaic:
>I'm a bit worried about using seq. For example, wouldn't the Seq be enumerated at least twice [..]

I was going back and forth on that one, to be honest.

 * Should the algorithm be [optimized upfront](http://c2.com/cgi/wiki?PrematureOptimization)?
 * Do locally copied, mutated, arrays make the code easier to maintain?

tests/FsCheck.Test/Gen.fs
moodmosaic: @kurtschelfthout It *might* be possible to strengthen that test, with the risk of turning it into [an erratic one](http://xunitpatterns.com/Erratic%20Test.html), because of potential [false positives](http://xunitpatterns.com/false%20positive.html).

Taking an `NonEmptySet` to get rid of the precondition sounds like a good idea though :smiley:

moodmosaic: @ploeh True, but then by applying the Devil's Advocate technique, the implementation may as well be this?

```
let shuffle xs = xs
```

:confused:

src/FsCheck/Gen.fs
moodmosaic: The generator used in the [Haskell QuickCheck version of `shuffle`](https://github.com/nick8325/quickcheck/commit/6c360c1b594b0243e51bfcd4e2428882570e7a51) is

```
choose (minBound :: Int, maxBound)
```

Where `mindBound :: Int` and `maxBound :: Int` values are:

```
λ> minBound :: Int
-9223372036854775808

λ> maxBound :: Int
9223372036854775807
```

---

If I use something like `int.MinValue` and `int.MaxValue` the Random algorithm throws an exception:

```
---- System.DivideByZeroException : Attempted to divide by zero.

----- Inner Stack Trace -----
..src\FsCheck\Random.fs(82,0): at FsCheck.Random.stdRange(Int32 l, Int32 h, StdGen rng)
..src\FsCheck\Random.fs(84,0): at FsCheck.Random.range@84-1.Invoke(StdGen rng)
..src\FsCheck\Gen.fs(163,0): at FsCheck.Gen.Choose@163.Invoke(Int32 _arg1, StdGen r)
..src\FsCheck\Gen.fs(295,0): at FsCheck.Gen.go@290-1T.Invoke(FSharpList`1 gs, ..)
..src\FsCheck\Gen.fs(63,0): at FsCheck.GenBuilder.bind@62.Invoke(Int32 n, StdGen r0)
..src\FsCheck\Gen.fs(157,0): at FsCheck.Gen.sample@156.Invoke(Int32 i, StdGen seed, ..)
..tests\FsCheck.Test\Helpers.fs(11,0): at FsCheck.Test.Helpers.sample@11.Invoke(Gen`1 gn)
..tests\FsCheck.Test\Gen.fs(126,0):
    at FsCheck.Test.Gen.Shuffle@125.Invoke(Unit unitVar)
    at System.Lazy`1.CreateValue()
    at System.Lazy`1.LazyInitValue()
    at System.Lazy`1.get_Value()
..src\FsCheck\Testable.fs(145,0): at FsCheck.Testable.Prop.safeForce[a](Lazy`1 body)
```

Perhaps this is a bug with the current Random implementation.

tests/FsCheck.Test/Gen.fs
ploeh: Yes, it could be, but since _shuffle_ is essentially a randomising operation, I can't think of any other property that will _always_ hold.

It's possible that a shuffle by chance will order the elements in the original order, unless, of course, you specify that a shuffle operation must always reorder at least two elements.

src/FsCheck/Gen.fs
ploeh: If a person comes by this code later on, and attempts to understand it, and has access to only the code, but not this discussion thread, would he or she be any wiser?

moodmosaic: True, these magic numbers result to a high(er) cognitive load. They could as well be bound to names, hypothetically `minBound` and `maxBound`.

tests/FsCheck.Test/Gen.fs
moodmosaic: I think the test, the way it currently is, should be pretty robust:

```fsharp
// Create the generator
Gen.shuffle xs
// Run the generator 10 times, yielding 10 different shuffled
// outputs based on the input.
|> sample 10
// Even if by chance the elements are ordered in the original
// order, there is at least _one_ from the 10 outputs that _is_
// shuffled.
|> List.exists ((<>) (Set.toSeq xs))
```

src/FsCheck/Gen.fs
```diff
@@ -313,6 +313,20 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    ///Generates a random permutation of the given sequence.
+    //[category: Creating generators]
+    [<CompiledName("Shuffle")>]
+    let shuffle xs =
+        gen {
+            let h = 922337203
```
ploeh: I'm sorry to beat a seemingly dead horse, but as a reader of this code, imagining that I have no prior knowledge of our conversation, I'd still wonder what's so special about this particular number...

moodmosaic: How about `bounds` or `max` (?)

kurtschelfthout: Not being able to use int.MinValue and int.MaxValue is indeed a bug in the random implementation. As you know I intend to replace it for v3.0...

kurtschelfthout:
Ok, you both mentioned or alluded to premature optimization! That's my pet peeve. :)

My sincere apologies beforehand for the rant...

Notice he said: "We should forget about **small** efficiencies, say about 97% of the time" (emphasis mine).

My concern is this code is at least twice as slow as it needs to be, because the seq is enumerated twice. How long do you think this should take?

```
Gen.shuffle (seq { for i in 1..10 do System.Threading.Thread.Sleep(100); yield i })|> Gen.sample 1 1 |> ignore
```
I would reasonably expect 1 second. Yet it takes 2.

Then, there is a known algorithm available for this which is linear: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle 

I love Haskell, but a problem it seems to have is that most people optimize for nice looking code at the expense of all else. We end up with one-liners that purport to be an implementation of a well-known algorithm but end up as something else: http://www.informit.com/articles/article.aspx?p=1407357&seqNum=3 This often happens with in-place algorithms...

Anyway, here's an implementation using Fisher-Yates or Knuth shuffle:

```
let shuffle2 xs =
        //Do the Fisher-yates shuffle.
        let xs = xs |> Seq.toArray
        let swap (arr:_ array) i j  =
            let v = arr.[j]
            arr.[j] <- arr.[i]
            arr.[i] <- v
        gen {
            let copy = Array.copy xs
            let maxI = copy.Length-1
            let! indexes = [ for i in 1..maxI-1 -> choose(i,maxI) ] |> sequence
            indexes |> List.iteri (swap copy)
            return copy
        }
```

I modified your original implementation to add Seq.toArray at the end (otherwise I think the comparison is unfair...and for reasons mentioned above I believe we need something like that anyway):
```
let shuffle xs =
        gen {
            let g = choose (-922337203, 922337203)
            let! ns = listOfLength (Seq.length xs) g
            return xs
                   |> Seq.zip ns
                   |> Seq.sortBy fst
                   |> Seq.map snd
                   |> Seq.toArray
        }
```
Some numbers in 64bit FSI with a Release build:

```
let s1 = Gen.shuffle  [1..10000]
let s2 = Gen.shuffle2 [1..10000]

> s1 |> Gen.sample 1 1000 |> ignore
Real: 00:00:07.681, CPU: 00:00:07.690, GC gen0: 761, gen1: 240, gen2: 1
val it : unit = ()

> s2 |> Gen.sample 1 1000 |> ignore
Real: 00:00:05.193, CPU: 00:00:05.194, GC gen0: 719, gen1: 315, gen2: 1
val it : unit = ()
```
And this seems relatively consistent across the board.

Again, my apologies for the rant. I was triggered by the premature optimization quote, it is - in my opinion - often misunderstood.

tests/FsCheck.Test/Gen.fs
kurtschelfthout: Yes, fair point. I misunderstood - I was thinking we could check that all the shuffles are indeed shuffles (i.e. they don't add or remove or duplicate elements) but that's really a different test.

src/FsCheck/Gen.fs
ploeh: Does `bounds`or `max` communicate the _origin_ of the value itself?

Can that origin be communicate by the value's _name_, or would it be possible to communicate the information via other means?

ploeh:
> you both mentioned or alluded to premature optimization

I can see how my comment could be interpreted that way, but it wasn't meant that way. The point was that with any high-level language, attempting to predict [what performs better](https://ericlippert.com/2012/12/17/performance-rant/) is usually futile, because there are more variables in play than you can possibly hold in your head. These may include

- Compiler optimisations (that you may or may not be aware of; I'm usually not aware of them)
- JIT compilation optimisations
- The size of L1, L2, etcetera caches
- The contents of L1, L2, etcetera caches when your particular block of code executes
- The number of cores on the system
- The amount of memory in your system
- Other processes running
- and so on...

It wasn't my intention to pull the Knuth quote into this discussion, but I see that you _did_ perform some measurements, and that's cool :+1:

ploeh:
Here are measurements executed on my 3½-year old Lenovo X1 Carbon with an i5 CPU:

```
> s1 |> Gen.sample 1 1000 |> ignore;;
Real: 00:00:11.327, CPU: 00:00:11.328, GC gen0: 587, gen1: 274, gen2: 2
val it : unit = ()
> s2 |> Gen.sample 1 1000 |> ignore;;
Real: 00:00:09.275, CPU: 00:00:09.296, GC gen0: 460, gen1: 276, gen2: 1
val it : unit = ()
```

Although overall slower, it seems to replicate the measurements reported by @kurtschelfthout that `s2` is somewhat faster than `s1`.

ploeh:
Interestingly, I was just fiddling around with the pure implementation, and came up with this variation:

```F#
let shuffle xs =
    let a = xs |> Seq.toArray
    gen {
        let g = choose (-922337203, 922337203)
        let! ns = listOfLength (a.Length) g
        return a
                |> Seq.zip ns
                |> Seq.sortBy fst
                |> Seq.map snd
    }
```

Executing it on my machine, it runs like this:

```
> s1 |> Gen.sample 1 1000 |> ignore;;
Real: 00:00:09.734, CPU: 00:00:09.718, GC gen0: 386, gen1: 163, gen2: 0
val it : unit = ()
```

That's so close to `s2` on my machine that the difference may not be significant... It's ever so slightly slower, but on the other hand has fewer garbage collections...

kurtschelfthout:
@ploeh I think you forgot a `Seq.toArray` at the end there. With it, it's as slow as the original, but it does avoid the problem with enumerating the input seq twice.

moodmosaic:
@kurtschelfthout Isn't `Seq.toArray` at the end constraining the input of `shuffle`

```
val shuffle : xs:seq<'a> -> Gen<'a []>
```

?

moodmosaic:
No, it doesn't. Please ignore my previous comment.

moodmosaic:
Based on all the above, I can see at least 2 choices for going forward:

**Fisher–Yates based shuffle**

As proposed by @kurtschelfthout, formatted:

```fsharp
let shuffle xs = 
    let xs = xs |> Seq.toArray
        
    let swap (arr : _ array) i j = 
        let v = arr.[j]
        arr.[j] <- arr.[i]
        arr.[i] <- v
    gen { 
        let copy = Array.copy xs
        let maxI = copy.Length - 1
        let! indexes = [ for i in 1..maxI - 1 -> choose (i, maxI) ]
                        |> sequence
        indexes |> List.iteri (swap copy)
        return copy
    }

//val shuffle : xs:seq<'a> -> Gen<'a []>
```

This *may* be harder to follow, unless one knows about [the algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle), it uses mutation, **but** performs ~1.5x faster than the proposed `shuffle`.

**QuickCheck 2.8.2 based shuffle**

As proposed in this Pull Request, and also using @ploeh's suggestion to avoid enumerating the input seq twice:

```fsharp
let shuffle xs = 
    let xs = xs |> Seq.toArray
    gen { 
        let g = choose (-922337203, 922337203)
        let! ns = listOfLength (xs.Length) g
        return xs
                |> Seq.zip ns
                |> Seq.sortBy fst
                |> Seq.map snd
                |> Seq.toArray
    }

// val shuffle : xs:seq<'a> -> Gen<'a []>
```

This is slower, as shown in the performance measurements, and also uses magic numbers (for the upper and lower range where random integers are picked to shuffle the input sequence).

---

Given all that, it turns out that the Fisher–Yates shuffle looks more standarized, performs faster, and the mutation part of it is supported by the language itself, so it may not hurt that much.

I'd add an additional commit to change the `shuffle` with the Fisher–Yates one and also with a URL in the commit message pointing in this great discussion.

Thoughts?

tests/FsCheck.Test/Gen.fs
```diff
@@ -119,6 +119,12 @@ module Gen =
         Gen.listOfLength length (Gen.constant v)
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
+
+    [<Property>]
+    let Shuffle (NonEmptySet (xs:Set<int>)) =
+        Gen.shuffle xs
+        |> sample 10
+        |> List.exists ((<>) (Set.toSeq xs))
```
ploeh: This doesn't look correct to me.

As an example, couldn't `xs` conceivably be `set [0]`? If so, all shuffles are the same as the original input list:

```F#
> let xs = set [0];;

val xs : Set<int> = set [0]

> Gen.shuffle xs |> Gen.sample 1000 10;;
val it : seq<int> list =
  [seq [0]; seq [0]; seq [0]; seq [0]; seq [0]; seq [0]; seq [0]; seq [0];
   seq [0]; seq [0]]
```

Interestingly, the property still passes:

```F#
> let xs = set [0];;

val xs : Set<int> = set [0]

> Gen.shuffle xs |> Gen.sample 1000 10 |> List.exists ((<>) (Set.toSeq xs));;
val it : bool = true
```

Investigating further, so does a variation with `forall`:

```F#
> let xs = set [0];;

val xs : Set<int> = set [0]

> Gen.shuffle xs |> Gen.sample 1000 10 |> List.forall ((<>) (Set.toSeq xs));;
val it : bool = true
```

The reason, it seems, is that sequences don't have structural equality (neither would I expect them to):

```F#
> let xs = set [0];;

val xs : Set<int> = set [0]

> seq [0] = Set.toSeq xs;;
val it : bool = false
```

ploeh: Indeed, it turns out that if I strengthen the equality comparison to compare lists (which _do_ have structural equality) instead of sequences, the property ultimately fails:

```F#
[<Property(MaxTest = 100000)>]
let Shuffle (NonEmptySet (xs:Set<int>)) =
    let l = Set.toList xs
    Gen.shuffle l
    |> sample 10
    |> List.map Seq.toList
    |> List.exists ((<>) l)
```

This fails as expected:

```
Test 'FsCheck.Test.Gen.Shuffle' failed: FsCheck.Xunit.PropertyFailedException : 
Falsifiable, after 2 tests (0 shrinks) (StdGen (2120765865,296145085)):
Original:
NonEmptySet (set [0])
```

ploeh:
If I take the proposed Fischer-Yates implementation, and attempt to exercise it with some boundary conditions, it behaves surprisingly:

```
shuffle List.empty<int> |> sample 1000 10;;
val it : int [] list =
  [[||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]]
> shuffle [0] |> sample 1000 10;;
val it : int [] list =
  [[|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]]
> shuffle [0; 1] |> sample 1000 10;;
val it : int [] list =
  [[|0; 1|]; [|0; 1|]; [|0; 1|]; [|0; 1|]; [|0; 1|]; [|0; 1|]; [|0; 1|];
   [|0; 1|]; [|0; 1|]; [|0; 1|]]
```

Shuffling the empty list, and a singleton seem OK, but the results of shuffling a two-element list looks suspicious.

Using the pure implementation looks less surprising in my opinion:

```
> shuffle List.empty<int> |> sample 1000 10;;
val it : int [] list =
  [[||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]; [||]]
> shuffle [0] |> sample 1000 10;;
val it : int [] list =
  [[|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]; [|0|]]
> shuffle [0; 1] |> sample 1000 10;;
val it : int [] list =
  [[|0; 1|]; [|0; 1|]; [|0; 1|]; [|0; 1|]; [|1; 0|]; [|0; 1|]; [|1; 0|];
   [|0; 1|]; [|1; 0|]; [|0; 1|]]
```

kurtschelfthout:
I love this PR :)

@moodmosaic yes, agreed - I have a preference for Fisher-Yates shuffle. We can put a link to a description of the algorithm in the code as well.

@ploeh There are three things that are difficult in computer science: cache invalidation and off by one errors. Here is a correcter version:

``let shuffle2 xs =
        //Do the Fisher-yates shuffle.
        let xs = xs |> Seq.toArray
        let swap (arr:_ array) i j  =
            let v = arr.[j]
            arr.[j] <- arr.[i]
            arr.[i] <- v
        gen {
            let copy = Array.copy xs
            let maxI = copy.Length-1
            let! indexes = [ for i in 0..maxI-1 -> choose(i,maxI) ] |> sequence
            indexes |> List.iteri (swap copy)
            return copy
        }
```

Nice find!

File: tests/FsCheck.Test/Gen.fs (hunk 4 above)
Comment (moodmosaic) [inline tests/FsCheck.Test/Gen.fs]: >sequences don't have structural equality (neither would I expect them to)

Same here :confused:

I've observed the test failing in the beginning, by taking a list as input, then started changing the test - even though [I shouldn't do that](http://blog.ploeh.dk/2013/04/02/why-trust-tests/) - and then it eventually became less trustworthy.

Now the question is, how should we *verify* this beast...

Comment (ploeh) [inline tests/FsCheck.Test/Gen.fs]: > how should we verify this beast

As previously outlined, I can only come up with one property that always holds (and that is that if you sort both the original list and the shuffled list, the sorted lists should be equal to each other). Well, I can think of other properties, such as the original and shuffled list should have the same length, but such a property is 'contained' in that other property...

It's true that the property that I propose doesn't _guarantee_ that a sequence is shuffled, because a degenerate implementation could pass the test simply by returning the input sequence untouched.

All other tests I can think of rely on random behaviour. I can only think of two ways to test with randomness:

- Use deterministic random number generators. If you use a deterministic random number generator (with a constant seed), determinism is assured. Using the _Devil's Advocate_ technique, however, a malign implementer could hide in the shadows of such tests, though.
- Rely on statistics and write tests that can, but are _extraordinarily_ unlikely to, produce false positives.

If there are more alternatives, I'd love to learn them :smile:

Comment (Aaronontheweb) [inline tests/FsCheck.Test/Gen.fs]: No need for RNG. Just use a standard Shuffle algorithm, like [Fisher-Yates](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle)

I implemented an example here:

https://github.com/Aaronontheweb/faker-csharp/blob/v2.0/Faker/Helpers/ArrayHelpers.cs#L61

Model tests for said function: shuffled order should never match original https://github.com/Aaronontheweb/faker-csharp/blob/v2.0/Faker.Models.Tests/ArrayHelperSpecs.cs#L12

I also wrote a composite test that verifies that the shuffled sequence always contains the same original elements

https://github.com/Aaronontheweb/faker-csharp/blob/v2.0/Faker.Models.Tests/TypeTableSpecs.cs#L62

Comment (Aaronontheweb) [inline tests/FsCheck.Test/Gen.fs]: Ah whoops, just saw that you did that below

Comment (ploeh) [inline tests/FsCheck.Test/Gen.fs]: > shuffled order should never match original

So, you're saying that if the input is e.g. `[0; 1]` then the output _must_ be `[1; 0]`, and must never be `[0; 1]`?

Comment (kurtschelfthout) [inline tests/FsCheck.Test/Gen.fs]: @ploeh no that isn't right. It should produce a uniform distribution across all possible permutations of the input, including itself. In other words, `[0;1]` or `[1;0]` must be equally likely.

I think we just check 1) that it is a permutation and 2) that if we take a list of length 5 and produce  10 permutations (i.e. sample 10), that at least one of those permutations is different from any of the others.

The chance of this *not* holding in case of a uniform shuffle - in other words, the chance of producing 10 times the same input list of length 5 - is (1/5!)^10, which is small enough. (unless my probability is off, which it may very well be :) )

Comment (ploeh) [inline tests/FsCheck.Test/Gen.fs]: > no that isn't right. It should produce a uniform distribution across all possible permutations of the input, including itself. In other words, [0;1] or [1;0] must be equally likely.

That's my interpretation of Fisher-Yates as well :+1: 

Relying on the statistical attribute of a test being exceedingly unlikely to produce a false positive is, I agree, sometimes the only option. I have tests like that in other code bases as well.

For the sake of someone not aware of this discussion thread, however, I think it might be a good idea to leave a comment in the test, to the effect that this property is non-deterministic, and may fail once in a blue moon.

In this way, if anyone clones the repository, runs all tests, and is unlucky enough to witness a false positive, at least there's 'documentation' in place stating that this can happen, and that one should simply run the tests again.

Comment (moodmosaic) [inline tests/FsCheck.Test/Gen.fs]: That looks like a good idea. I'll do that :+1:

**Comment by moodmosaic** [comment]
I've just added a9af0ae switching to the Fisher–Yates shuffle.

**Comment by kurtschelfthout** [comment]
Awesome work @moodmosaic. This looks ready to go, would you mind just updating your branch to latest from master?

**Comment by moodmosaic** [comment]
@kurtschelfthout Done.

**Comment by moodmosaic** [comment]
Thank you everyone for the _awesome_ feedback!

File: src/FsCheck/Gen.fs [hunk 5]
```diff
@@ -313,6 +313,24 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOfLength n arb = sequence [ for _ in 1..n -> arb ]
 
+    ///Generates a random permutation of the given sequence using the Fisher-Yates shuffle algorithm.
```
ploeh: Isn't the specific algorithm used an implementation detail? Doesn't this documentation currently promise to always use that algorithm? What if, in the future, the need arises to switch to a different algorithm?

moodmosaic: It shouldn't be a breaking change, as long as `shuffle` remains the same

```
val shuffle : xs:seq<'a> -> Gen<'a []>
```

and as long as it's properties also remain the same.

That comment is there in order to, hopefully, help any code reader understand easier what's going on, if he or she is not aware of that particular algorithm.

Could we do better?

src/FsCheck/Gen.fs
```diff
@@ -313,6 +313,24 @@ module Gen =
...
+    let shuffle xs =
+        let xs = xs |> Seq.toArray
+        let swap (arr : _ array) i j =
+            let v = arr.[j]
+            arr.[j] <- arr.[i]
+            arr.[i] <- v
+        gen {
+            let copy = Array.copy xs
+            let maxI = copy.Length - 1
+            let! indexes = [ for i in 0..maxI - 1 -> choose (i, maxI) ]
```
ploeh: Why `maxI - 1`?

kurtschelfthout: Because the last index `maxI` you don't need to swap with itself.

src/FsCheck/Gen.fs
kurtschelfthout: Apologies,  I missed these comments before I merged. Happy to take further improvements as further PRs though.

ploeh: In my opinion, if you state in a piece of documentation that a function has a certain quality, then that's a guarantee you make. It would be a breaking change to subsequently change this. 

That's my position on the matter, but @kurtschelfthout may be of a different opinion.

src/FsCheck/Gen.fs
ploeh: Wasn't this the off-by-one-bug we previously discussed? 

(Sorry to ask instead of investigating myself, but I'm on a mobile phone with only intermittent connection most of the day today, so it's a bit unwieldy to look up and experiment with branches and so on.)

src/FsCheck/Gen.fs
moodmosaic: I'm happy to send a PR that puts the comment back to how it was. We may though consider putting a link to a description of the algorithm in the code, as @kurtschelfthout suggested [here](https://github.com/fscheck/FsCheck/pull/221#issuecomment-211901685).

src/FsCheck/Gen.fs
kurtschelfthout: No, the off by one was I started at 1, i.e. `in [1..maxI-1]`.

One way to read that line is it determines the target index to swap with for each original index `i`. For `i = maxI` the `choose` is `choose(maxI,maxI)` which produces `maxI`.

src/FsCheck/Gen.fs
kurtschelfthout: I think it's a fair point, if it's in the public doc changing it is a breaking change of sorts. The algorithm is more of an implementation note, happy to have links there.

PR #227, opened 2016-04-21
Added a property about Gen.shuffle.

ploeh:
With the previous test suite, I could use the Devil's Advocate review
technique to reduce Gen.shuffle to this:

```F#
let shuffle (xs : 'a seq) = gen { return Array.empty<'a> }
```

Despite this degenerate, and obviously incorrect, implementation, all
tests passed. This demonstrated the need for more tests, so I added one.

I'm not 100% certain that the term 'isomorphic' entirely covers this
property, but if we regard the domain as an ordered collection of a
certain size, and the codomain as equal to the domain, then shuffle is
isomorphic because every element in the input is mapped to a position in
the output, and a reverse map exists.

### DIFF
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -120,6 +120,13 @@ module Gen =
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
 
+    [<Property>]
+    let ``shuffle is isomorphic`` (xs : int array) =
+        Gen.shuffle xs
+        |> sample1
+        |> Array.sort
+        |> ((=) (Array.sort xs))
+
     // This property is non-deterministic, and may rarely fail. The chance of
     // this property not holding in case of a uniform shuffle - in other words,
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Nice. 

Re: test name what do you think about `shuffle generates a permutation` or something along those lines?

tests/FsCheck.Test/Gen.fs
```diff
@@ -120,6 +120,13 @@ module Gen =
         |> sample1
         |> ((=) (List.init length (fun _ -> v)))
 
+    [<Property>]
+    let ``shuffle is isomorphic`` (xs : int array) =
+        Gen.shuffle xs
+        |> sample1
+        |> Array.sort
+        |> ((=) (Array.sort xs))
+
```
moodmosaic: :+1: 

It looks indeed like both input and output are isomorphic, given than both are of the same form and shape, and there's a structure-preserving one-to-one correspondence between them.

ploeh:
Good name suggestion :+1: Pull request updated.

PR #232, opened 2016-04-26
Include Gen.growingElements in the docs

moodmosaic:
This PR adds `growingElements` to the list, under the *Useful Generator Combinators* section of the docs.


kurtschelfthout:
There's no need to update the .md file, it's no longer used. :) I will remove it.

Issue #233, opened 2016-04-28
Experimental.StateMachine does not shrink Operations


MarcoMaar:
As mentioned in  7e62802 the Shrinker of StateMachine only shrinks the list of Operations but not the Operations themselves. I create a PR for further discussion.

MarcoMaar:
closed with d57d928

PR #234, opened 2016-04-28
StateMachine Shrinking shrinks Operations

MarcoMaar:
Addresses #233

There are still two open issues:
1.) I am not entirely sure how I could write a testcase that checks if the Shrunk value is correct(3) and not 5. Of course it is possible to catch the exception and check the exception message but there should be a better way. Currently the testcase in the PR is just failing, so that needs to be fixed.

2.) Since I am using the defaultListShrinker again, I am not sure which implications this has. The testcase ``shrinker should remove loops`` still passes. Now and before it was not possible to have a transition sequence like [a->b;b->c] with an error in the state c shrunk to [a->c] in this testcase.

### DIFF
--- a/src/FsCheck/StateMachine.fs
+++ b/src/FsCheck/StateMachine.fs
@@ -294,15 +294,14 @@ module StateMachine =
 //                seq { for i in 1..l.Length-1 do
 //                        yield! Seq.windowed i l //|> Seq.map Seq.toList
 //                }
-//
+////
 //            allSubsequences l |> Seq.distinct
 //
 //            let skipOne (l:list<_>) =
 //                seq { for i in 0..l.Length-1 do 
 //                        yield List.foldBack (fun e (c,r) -> c+1, if i <> c then e::r else r) l (0,[]) |> snd
-//                }
-            let defaultListShrinker = Arb.Default.FsList().Shrinker l
-            defaultListShrinker
+//                    }
+            l |> Arb.Default.FsList().Shrinker
 
         run.Operations 
         |> List.map fst
--- a/src/FsCheck/StateMachine.fs
+++ b/src/FsCheck/StateMachine.fs
@@ -2,7 +2,7 @@
 **  FsCheck                                                                 **
 **  Copyright (c) 2008-2015 Kurt Schelfthout and contributors.              **
 **  All rights reserved.                                                    **
-**  https://github.com/fscheck/FsCheck                              **
+**  https://github.com/fscheck/FsCheck                                      **
 **                                                                          **
 **  This software is released under the terms of the Revised BSD License.   **
 **  See the file License.txt for the full text.                             **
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
1) You can create a sequence that contains an operation and run it through the shrinker, and see if it produces what you think it should shrink to.
2) I think it just tries more possibilities that may not make sense. So it may have a perf impact. Would have to check the list shrinker to make sure.

Isn't it easier overall to add srhinking of the operations to the list of candidates, like here: https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Arbitrary.fs#L472

MarcoMaar:
1) oh yeah.. stupid of me to not realize... you even used it in the test case I described. Will fix that.
2) I did not think about performance, makes sense. Okay, so I will not use FsList() shrinker but the line used in it with the operation shrinker.

MarcoMaar:
1) should work
2) I checked the `FsList().Shrinker` and I think it operates quite similar to the `operationShrinker`. Both skip an element in the list but `FsList().Shrinker` also shrinks the elements itself. Isn't that what we want?

kurtschelfthout:
Maybe, the list shrinker certainly does more:
https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Arbitrary.fs#L471

Due to the recursive call to shrink it also skips two elements, three etc. But the statemachine specific shrinker already removes subsequent operations for which the precondition now fails and keeps the rest.

MarcoMaar:
That is what I thought first as well. But I tried out some values and saw that it does not behave that way. I took my time to look at it a bit more and I think I found the reason.

for the current version this property passes
```fsharp
    let ``length of shrunk list is at minimum n-1 of list``() =
        let prop l = 
            let shrunk = Arb.Default.FsList().Shrinker l
            l.Length = 0 || (shrunk |> Seq.forall (fun s -> s.Length = l.Length || s.Length = l.Length - 1))
        Check.QuickThrowOnFailure prop
```
which means it does not skip more than one element.
The reason is the line you pointed out 
https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Arbitrary.fs#L471

yield only returns one element therefore it should use yield bang I think. e.g. change the line to this
`for xs' in shrink xs do yield! __.Shrinker xs'`
and now the property fails because it generates smaller lists as well.

So which behavior do we want for the FsList shrinker? The one that skips 1 element or the one that can skip more than one element. I assume for operations we want the one that skips 1 element. Both versions also shrink the elements in the list.

It might be a bit tricky to write a test case(property) that can distinguish between the two versions.

kurtschelfthout:
Good point, I misread. Ok, let's keep the list shrinker as is, seems to work well enough, and use it here too.

MarcoMaar:
1)but it still steps into the recursion all the way to `Seq.empty`. Shouldn't we fix that for performance reasons?

2)and should we add the testcase to the repo, to ensure the behavior?

3)At least we should change the xml comment of FsList because it is misleading.
`        ///Generate a list of values. The size of the list is between 0 and the test size + 1.`

Sorry for so many questions, I just want to make sure before I upload unnecessary code.

kurtschelfthout:
1) Maybe, if we can show it's indeed a performance problem. In any case the list has to be catenated, so it may not matter.
2) Sure, could add a test here: https://github.com/fscheck/FsCheck/blob/master/tests/FsCheck.Test/Arbitrary.fs
3) How is that misleading?

MarcoMaar:
1) okay, I will check the performance.
2) cool, will do that
3) sorry, my mistake. I was thinking about the shrinker not about the arbitrary.

Issue #218, opened 2016-04-17
Add samples to Generator and Shrinker documentation


kurtschelfthout:
While talking to people about FsCheck I had the following idea to improve the documentation.

The documentation page https://fscheck.github.io/FsCheck/TestData.html which is generated from https://github.com/fscheck/FsCheck/blob/master/docs/content/TestData.fsx currently feels very abstract - I think it would be improved if some of the generators and shrinkers discussed would show some actual output. This would also show people how to write/test generators/shrinkers and get a feel for what they do.

Concretely, I would suggest to add executed code samples (using FSharp.Formatting's output capturing, like in the other docs) that essentially call `Gen.sample` and the shrinker function with an example value.

kurtschelfthout:
Awesome work so far. I have pushed the updated docs.

ploeh:
With #266 we've now covered all the combinators under the _Useful Generator Combinators_ with the exception of `Gen.tryFilter`/`Gen.tryWhere`. I'll be happy to _write_ something about that as well, but I can't think of a simple example :flushed: So far, I've never had the need to use this function.

Can anyone think of a _simple, reasonable_ example?

kurtschelfthout:
Nice.

`tryWhere` is used in FsCheck itself to try and generate the next operation in a sequence of operations that satisfy a user-provided precondition. If it can't find any, it stops the sequence. 

Perhaps an example where we are building a list or data structure, where with each added element the chance of finding another element with that satisfies some global invariant becomes smaller. If we get `None` from `tryWhere` we know we have tried "hard", and either give up and return where we are so far (if we are incrementally generating the data structure, i.e. adding one element with each recursive step, for example) or try to generate a smaller data structure (thus hopefully making the global invariant easier to satisfy)

E.g. concretely generate a list of distinct elements from a given generator  - where the generator has a generic type, so we don't know much about how easy it is to generate a list of district elements, e.g. compare a list of  distinct bools vs distinct floats. We can do this by calling `Gen.listOfLength` with the size, then using `Gen.tryWhere` to check if the elements are indeed distinct. If that returns `None`, we should try with a smaller length of list. If it returns `Some l` we can return `l`.

moodmosaic:
Should we document `Gen.eval` also? It's a bit low-level though...

If we should document it, I'd be happy to open a PR.

ploeh:
Would you ever use `Gen.eval` when writing properties?

moodmosaic:
[...]

PR #235, opened 2016-04-30
Documented Gen.elements by describing it

ploeh:
and providing examples.

This pull request is an attempt at (a start of) addressing #218. I'd like to add documentation and examples for other combinators, like `listOf`, `two`, etcetera, but I wanted to start with a single entry in case this wasn't quite what you had in mind.

I chose to start with `Gen.elements` because I find it a useful foundation on which to build other examples. As an example, once the reader understands `Gen.elements`, hopefully this example of `Gen.two` should be easy to understand:

```F#
> Gen.elements [0..9] |> Gen.two |> Gen.sample 1000 10 |> Seq.toList;;
val it : (int * int) list =
  [(7, 9); (0, 2); (5, 9); (8, 4); (3, 7); (3, 2); (1, 2); (9, 0); (8, 8);
   (0, 5)]
```

### DIFF
--- a/docs/content/TestData.fsx
+++ b/docs/content/TestData.fsx
@@ -168,6 +168,54 @@ If `g` is a generator for type `t`, then
 - If xs is a sequence, then `shuffle xs` generates a random permutation of xs.
 
 All the generator combinators are functions on the Gen module. In C#, the names are the same just capitalized differently.
+
+### Generator examples
+
+The following examples use `Gen.sample` in order to show example output. In
+general, you shouldn't use `Gen.sample` when writing properties, but it can be
+helpful when developing or troubleshooting a useful custom generator.
+
+Please be aware that due to the non-deterministic nature of FsCheck, the output
+of calling `Gen.sample` will, in most cases, differ between calls.
+
+Most examples below use `Gen.sample` to generate a small list of example
+values, for example a list of ten generated values.
+
+#### Elements
+
+You can use the `Gen.elements` function to create a generator of singular
+values drawn from a collection of possible values. The collection is inclusive,
+which means that both the first and last element, as well as all elements
+between, can be drawn.
+
+In the following example, the numbers from 0 to 9 define the collection of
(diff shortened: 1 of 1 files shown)


docs/content/TestData.fsx
```diff
@@ -168,6 +168,54 @@ If `g` is a generator for type `t`, then
...
+values drawn from a collection of possible values. The collection is inclusive,
+which means that both the first and last element, as well as all elements
+between, can be drawn.
+
+In the following example, the numbers from 0 to 9 define the collection of
+possible values. The result is a generator that creates `int` values guaranteed
+to be between 0 and 9 (both included).*)
+
+(***define-output:ElementsExample***)
+Gen.elements [0..9] |> Gen.sample 1000 10 |> Seq.toList
```
moodmosaic: I could imagine that some readers may wonder what's so special about `1000`.

docs/content/TestData.fsx
```diff
@@ -168,6 +168,54 @@ If `g` is a generator for type `t`, then
...
+a single integer drawn from the collection of numbers from 0 to 9:*)
+
+(***include-it:ElementsExample***)
+
+(**All elements are equally likely to be drawn from the collection; we say that
+the random function has a uniform distribution. One easy way to affect the
+distribution is to put more than one identical element into the collection:*)
+
+(***define-output:SkewedElementsExample***)
+Gen.elements ["foo"; "foo"; "bar"] |> Gen.sample 1000 10 |> Seq.toList
```
moodmosaic: [Same](https://github.com/fscheck/FsCheck/pull/235/files#r61679450) also here as well.

moodmosaic:
This is, IMHO, great work!

Right now, the size being passed in `Gen.sample` can be confusing. So, I'd suggest that we consider documenting the difference between the generators that take size into account and those that don't.

kurtschelfthout:
I think this is a great start. I am not so worried about the size aspect. This should become clear in time.

PR #237, opened 2016-05-02
FsList shrinker

MarcoMaar:
Following PR for #234

I did a performance comparison for three versions of the FsList shrinker:
1) old one
2)
```fsharp
        static member FsList() = 
            { new Arbitrary<list<'a>>() with
                override __.Generator = Gen.listOf generate
                override __.Shrinker (l:'a list) =
                    let skipOne (l:list<_>) =
                        seq { for i in 0..l.Length-1 do 
                              yield List.foldBack (fun e (c,r) -> c+1, if i <> c then e::r else r) l (0,[]) |> snd
                            }
                    let rec shrinks cl pre = 
                        match cl with
                        | [] -> Seq.empty
                        | [x] -> seq { for x' in shrink x -> x'::pre |> List.rev}
                        | x::xs -> seq { for x' in shrink x -> pre @ (x'::xs)
                                         yield! shrinks xs (x::pre |> List.rev) }
                    seq { yield! skipOne l; yield! shrinks l []}
```
3) the one in this PR

I did the performance check with three configurations. The small configuration uses startsize 0 and endsize 100 medium uses 200,200 and large uses 400,400. I ran each test 5 times and used the average. I did not use a stopwatch and simply looked at the time a testcase needed. The small configurations were quite similar due to the startup time of VS's test runner.
```fsharp
    [<Fact>]
    let ``performance_medium``() =
        let config = { Config.Quick with StartSize = 200; EndSize = 200 }
        let prop l = 
            let shrunk = Arb.Default.FsList().Shrinker l
            l.Length = 0 || (shrunk |> Seq.forall (fun s -> s.Length = l.Length || s.Length = l.Length - 1))
        Check.One(config, prop) 
```
Here are the results:
-- | ---S--- | --M-- | --L--
1 | 370ms | 5.5s | 40s
2 | 370ms | 2.8s | 19s
3 | 350ms | 2.0s | 14s

There is a difference between 3 and 1,2 we should consider. The shrunk values that skip one element are at the beginning of the sequence in 1 and 2. They are not in 3. It may make sense to use Implementation 2 if we want to have the skips at the beginning.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -464,12 +464,15 @@ module Arb =
         static member FsList() = 
             { new Arbitrary<list<'a>>() with
                 override __.Generator = Gen.listOf generate
-                override __.Shrinker l =
-                    match l with
-                    | [] ->         Seq.empty
-                    | (x::xs) ->    seq { yield xs
-                                          for xs' in shrink xs -> x::xs'
-                                          for x' in shrink x -> x'::xs }
+                override __.Shrinker (l:'a list) =
+                    let rec inner cl pre = 
+                        match cl with
+                        | [] -> Seq.empty
+                        | [x] -> seq { yield pre; for x' in shrink x -> x'::pre |> List.rev}
+                        | x::xs -> seq { yield pre @ xs //skip current element
+                                         for x' in shrink x -> pre @ (x'::xs) //shrink current element
+                                         yield! inner xs (x::pre |> List.rev) }
+                    inner l []
             }
 
         ///Generate an object - a boxed char, string or boolean value.
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -464,15 +464,12 @@ module Arb =
         static member FsList() = 
             { new Arbitrary<list<'a>>() with
                 override __.Generator = Gen.listOf generate
-                override __.Shrinker (l:'a list) =
-                    let rec inner cl pre = 
-                        match cl with
-                        | [] -> Seq.empty
-                        | [x] -> seq { yield pre; for x' in shrink x -> x'::pre |> List.rev}
-                        | x::xs -> seq { yield pre @ xs //skip current element
-                                         for x' in shrink x -> pre @ (x'::xs) //shrink current element
-                                         yield! inner xs (x::pre |> List.rev) }
-                    inner l []
+                override __.Shrinker l =
+                    match l with
+                    | [] ->         Seq.empty
+                    | (x::xs) ->    seq { yield xs
+                                          for xs' in shrink xs -> x::xs'
+                                          for x' in shrink x -> x'::xs }
             }
 
         ///Generate an object - a boxed char, string or boolean value.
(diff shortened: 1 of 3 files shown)


kurtschelfthout:
Cool. Initial impressions for now.

I prefer not to change the semantics  - although we don't have a test case for it, it may affect the eventual counter-example in some cases and I consider that a "slightly breaking" change ;) 

Also the perf difference between 2 and 3 seems pretty small. Shrinking is not a frequent operation so doesn't look like it would be really worth it. (As a tangent, the real meat of where perf matters I think is in generation.)

I'll try to take a look at this in more detail soon...

MarcoMaar:
> I prefer not to change the semantics - although we don't have a test case for it, it may affect the eventual counter-example in some cases and I consider that a "slightly breaking" change ;) 

You are totally right. So maybe the 2. implementation makes more sense than the 3.

Oh.. and since I am quite new to F# I wouldn't be surprised if there are more elegant and faster ways to do the same..

> I'll try to take a look at this in more detail soon...

Cool, let me know

kurtschelfthout:
So I ran a few tests with the following functions:

```
let rec shrink1 l =
     match l with
     | [] ->         Seq.empty
     | (x::xs) ->    seq { yield xs
                           for xs' in shrink1 xs -> x::xs'
                           for x' in Arb.shrink x -> x'::xs }

let shrink2 l =
    let skipOne (l:list<_>) =
        seq { for i in 0..l.Length-1 do 
                yield List.foldBack (fun e (c,r) -> c+1, if i <> c then e::r else r) l (0,[]) |> snd
            }
    let rec shrinks cl pre = 
        match cl with
        | [] -> Seq.empty
        | [x] -> seq { for x' in Arb.shrink x -> x'::pre |> List.rev}
        | x::xs -> seq { for x' in Arb.shrink x -> pre @ (x'::xs)
                         yield! shrinks xs (x::pre |> List.rev) }
    seq { yield! skipOne l; yield! shrinks l []}

let rec shrink3 l =
    let rec inner cl pre = 
        match cl with
        | [] -> Seq.empty
        | [x] -> seq { yield pre; for x' in Arb.shrink x -> x'::pre |> List.rev}
        | x::xs -> seq { yield pre @ xs //skip current element
                         for x' in Arb.shrink x -> pre @ (x'::xs) //shrink current element
                         yield! inner xs (x::pre |> List.rev) }
    inner l []
```

Which corresponds to your 1,2 and 3.

I then tried to shrink a list of 500 elements and time it:
```
let theList = [1..500]

shrink1 theList |> Seq.toArray |> ignore

shrink2 theList |> Seq.toArray |> ignore

shrink3 theList |> Seq.toArray |> ignore
```

I didn't see much difference between the first two; the third is faster but not dramatically so.

Also the order of the shrinks is different between the three approaches; it's not clear what you're trying to achieve here, is this a perf issue? Or is this a new shrinking strategy? In the latter case do you have a motivating example?

I do like the stronger test though.

MarcoMaar:
I was just investigating the performance, but I guess the faster implementation does not justify changing the order of shrinking.
So I will just commit the test case + removal of commented out code, right?

kurtschelfthout:
I see...Yes, removing comments and adding the stronger test is cool.

MarcoMaar:
can we merge this PR?

kurtschelfthout:
Sorry, didn't realize you had updated this.

MarcoMaar:
no problem (: 
thanks

Issue #236, opened 2016-05-01
Make Gen.sample easier to use in v3


moodmosaic:
Even though _not_ all generators depend on size, `Gen.sample` takes `size`, for setting the size.

``` f#
val sample : size:int -> n:int -> gn:Gen<'a> -> 'a list
```

Why not defining `Gen.sample` as:

``` f#
val sample : n:int -> gn:Gen<'a> -> 'a list
```

The size passed to the generator should always be the same; if one wants another size then he or she should explicitly use `Gen.resize`.

---

That's also how `sample` works in Erlang QuickCheck by QuviQ and Haskell QuickCheck.

moodmosaic:
Let me know if you'd want to see this in action on some v3-related branch.

kurtschelfthout:
I'm conflicted on this. I'm sympathetic to the argument that we have  `Gen.resize` to control this already. On the other hand we are back to picking a magic number for size now in `sample`.

moodmosaic:
>we are back to picking a magic number for size now in `sample`.

Not necessarily. — How about this:

```f#
let sample gn = [ for s in [ 0..2..20 ] -> resize s gn |> eval ]
```

(I haven't tried to compile this, but you should get the idea.)

kurtschelfthout:
Well, that's two magic numbers, `2` and `20` :)

moodmosaic:
👍

PR #238, opened 2016-05-02
Documented and provided examples for Gen.constant.

ploeh:
This is another pull request related to #218. I know I start with all the boring, simple stuff, but I think it's important to establish a foundation for more complex examples.

### DIFF
--- a/docs/content/TestData.fsx
+++ b/docs/content/TestData.fsx
@@ -181,6 +181,25 @@ of calling `Gen.sample` will, in most cases, differ between calls.
 Most examples below use `Gen.sample` to generate a small list of example
 values, for example a list of ten generated values.
 
+#### Constant
+
+The `Gen.constant` function is perhaps the simplest, and easiest, generator to
+understand. Even though it's part of a system that generates random values,
+this particular generator always returns the same value:*)
+
+(***define-output:ConstantExample***)
+Gen.constant (1, "Foo") |> Gen.sample 1000 10 |> Seq.toList
+
+(**In this example, the constant is a complex value (a tuple); it can also be a
+simple value, as for example as string or an integer. As you can see from the
+return value, all singular elements returned is the same tuple.*)
+
+(***include-it:ConstantExample***)
+
+(**Since the purpose of FsCheck is to generate random values, you shouldn't
+need to use `Gen.constant` often. Still, it can come in handy if you need to
+keep the value of a particular type constant while you vary other values.
+
 #### Elements
 
 You can use the `Gen.elements` function to create a generator of singular


kurtschelfthout:
Cool stuff. Maybe I'm starting to agree with @moodmosaic in that the `1000` size is a bit confusing. And maybe we shouldn't be repeating the `10` either. 

Perhaps for now it's a good idea to add a short, perhaps incomplete explanation at the beginning with a ref to where size is more properly explained, and define a helper function `let sample = Gen.sample 1000 10` (or however) at the beginning that is used where appropriate?

ploeh:
I agree that better explanation about the use of `Gen.sample` is warranted. As a start, I've submitted #240.

I'd also like to document the size number, but I'm not even sure that `1000` is the appropriate number. I simply selected it because `FsCheck.Test.Helpers.sample` use `1000` as the hard-coded size constant. To be honest, I find it a bit high, but I admit that my understanding of exactly how the various generators use the size is still hazy.

I'll be happy to add more information about that particular number (or another number), but for now, I don't know what to write, more than that it's a nice-looking, arbitrary number.

When it comes to defining a helper function, I'd suggest keeping each example as self-contained as possible. As the list of examples grow, a reader will have to jump around in the document in order to figure out how a hypothetical `sample` function is defined. If each example is self-contained, on the other hand, it's immediately clear what's going on, once you understand what `Gen.sample` does.

kurtschelfthout:
Perhaps you should use `-1` for the examples where the size doesn't matter to make things clear. I.e. all the examples you've done so far. :)

moodmosaic:
A negative size would make things more complicated. In v3, `sample` should better *not* take `size` at all, as proposed in #236.

ploeh:
> Perhaps you should use -1 for the examples where the size doesn't matter to make things clear.

Whatever makes it easier to understand the documentation, I'm all for it :smile: 

The difficult part of writing documentation, however, is that you're often blind while doing it. I don't know how to measure which way to present the information is best.

I'm not fixed on the number `10,000`, but I'm concerned that `-1` would be confusing. On the one hand, it does demonstrate that the number doesn't really matter. On the other hand, for someone not yet fully understanding what FsCheck is, it may be confusing: _how can_ size _be negative?!_

As @moodmosaic points out in #236 removing `size` from `Gen.sample` would make the API more intuitive (in my opinion), but I don't know whether or not it's feasible.

kurtschelfthout:
How about `0`? I'll reply on the other conversation about removing size from `sample`, but in any case that's a breaking change so it doesn't help us now.

My opinion on the whole size thing is that we explain it early or at least mention it as you've done. Sure there are a few of the simpler generator constructors that don't use size, but it's the minority. So as you've discovered with list, quite quickly you have to go into this topic anyway.

Distribution and size of generated values are important, too many people are overlooking it, and it's counter-productive imo to want to "hide" important things for newcomers because it puts them on the wrong track. Either they'll never understand and their use of FsCheck is severely crippled, or they'll be frustrated later when they find out that their nice property tests were in fact spending 99% of the time generating empty lists. At the risk of sounding grandiose, it's like trying to explain quantum mechanics without going into the particle/wave duality because it's "too confusing". You can't really understand things without understanding that.

PR #241, opened 2016-05-10
Changed Gen.listOf to stay below size in length.

ploeh:
Depending on your interpretation, this change could be considered a **breaking change**. Then again, perhaps it isn't :smile: Feel free to reject it if you disagree with this change. I actually started writing this up as an issue, but then went so deep into investigation and experimentation mode that I found it more appropriate to submit it as a pull request.

The previous version of `Gen.listOf` added 1 to the size in order to "decrease [the] chance of empty list". This had the peculiar effect that you'd see lists longer than `size`:

```
> [0..9] |> Gen.elements |> Gen.listOf |> Gen.sample 1 9;;
val it : int list list = [[4]; [9]; []; [7]; [0; 6]; [5]; [8]; [3]; []]
```

Notice the `[0; 6]` sample, which has a length of 2.

According to the documentation, this isn't wrong per se, since it only states that the "maximum length depends on the size parameter". Still, I found this unintuitive, and that it would make it harder to explain to beginners how FsCheck works.

For sizes from 1 to 10, the old implementation had these frequencies of empty lists:

```
Size  Empty lists
   1         42 %
   2         37 %
   3         32 %
   4         29 %
   5         26 %
   6         25 %
   7         23 %
   8         22 %
   9         21 %
  10         19 %
```

In the new implementation, I've fiddled a bit with the numbers to approximate something similar:

```
Size  Empty lists
   1         45 %
   2         37 %
   3         32 %
   4         29 %
   5         26 %
   6         24 %
   7         22 %
   8         21 %
   9         20 %
  10         18 %
```

As one can see, the new implementation creates empty lists with a slightly higher frequency for small sizes, but with lower frequencies as the size increases:

image

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -382,9 +382,11 @@ module Gen =
     //[category: Creating generators from generators]
     [<CompiledName("ListOf")>]
     let listOf gn =
-        sized <| fun n ->
-            gen {   let! k = choose (0,n+1) //decrease chance of empty list
-                    return! listOfLength k gn }
+        sized <| fun n -> gen {
+            let! k = choose (min 1 n, n)
+            return! frequency [
+                (2 + n, listOfLength k gn); // 2 + n seems to give a good distribution of non-empty
+                (1    , gen.Return []) ] }  // vs empty lists
 
     /// Generates a non-empty list of random length. The maximum length 
     /// depends on the size parameter.
--- a/tests/FsCheck.Test/Gen.fs
+++ b/tests/FsCheck.Test/Gen.fs
@@ -172,7 +172,7 @@ module Gen =
     let ListOf (NonNegativeInt size) (v:char) =
         Gen.resize size (Gen.listOf <| Gen.constant v)
         |> sample 10
-        |> List.forall (fun l -> l.Length <= size+1 && List.forall ((=) v) l)
+        |> List.forall (fun l -> l.Length <= size && List.forall ((=) v) l)
     
     [<Property>]
     let NonEmptyListOf (NonNegativeInt size) (v:string) =


src/FsCheck/Gen.fs
```diff
@@ -382,9 +382,11 @@ module Gen =
     //[category: Creating generators from generators]
     [<CompiledName("ListOf")>]
     let listOf gn =
-        sized <| fun n ->
-            gen {   let! k = choose (0,n+1) //decrease chance of empty list
```
moodmosaic: >choose (0, n+1) // decrease chance of empty list

I have a feeling that this shouldn't be the case anymore in the [replace-random](https://github.com/fscheck/FsCheck/tree/replace-random) branch.

In that branch the Hugs/Random's [StdGen](https://hackage.haskell.org/package/random-1.1/docs/System-Random.html) algorithm is replaced by [SplitMix](https://github.com/nkartashov/SplitMix/tree/master/src/System/Random).

kurtschelfthout:
This change makes the distribution of lengths not-uniform. I'm 80% convinced that is probably a bad idea...

src/FsCheck/Gen.fs
kurtschelfthout: No, don't think that affects it.

ploeh:
Is the distribution uniform now? Perhaps I'm missing something obvious, but it doesn't look uniform to me.

In order to investigate, I defined this function against the _current_ implementation of `Gen.listOf`:

```F#
let distributionBySize size =
    [0..9]
    |> Gen.elements
    |> Gen.listOf
    |> Gen.sample size 10000
    |> List.map List.length
    |> List.countBy id
    |> List.sortBy fst
```

This will tell you the count of each length, given a `size` value. Here are examples for sizes `1` and `2`:

```
> distributionBySize 1;;
val it : (int * int) list = [(0, 4193); (1, 4151); (2, 1656)]
> distributionBySize 2;;
val it : (int * int) list = [(0, 3592); (1, 3574); (2, 2021); (3, 813)]
```

That is, for `size 1`, there are 4,193 empty lists, 4,151 singleton lists, and 1,656 lists with two elements.

These don't look uniformly distributed to me. To illustrate, I plotted sizes 1 to 10 into a bar chart:

image

Additionally, even if you look at a size of `1000`, the distribution still looks skewed:

```
> distributionBySize 1000;;
val it : (int * int) list =
  [(0, 70); (1, 58); (2, 57); (3, 52); (4, 51); (5, 57); (6, 42); (7, 47);
   (8, 57); (9, 43); (10, 44); (11, 55); (12, 37); (13, 34); (14, 45);
   (15, 42); (16, 46); (17, 39); (18, 42); (19, 37); (20, 29); (21, 36);
   (22, 49); (23, 33); (24, 32); (25, 29); (26, 42); (27, 24); (28, 26);
   (29, 30); (30, 37); (31, 31); (32, 34); (33, 34); (34, 40); (35, 42);
   (36, 34); (37, 37); (38, 27); (39, 24); (40, 27); (41, 27); (42, 33);
   (43, 32); (44, 31); (45, 36); (46, 33); (47, 29); (48, 30); (49, 29);
   (50, 38); (51, 33); (52, 27); (53, 30); (54, 38); (55, 29); (56, 38);
   (57, 32); (58, 34); (59, 36); (60, 31); (61, 30); (62, 26); (63, 36);
   (64, 21); (65, 25); (66, 28); (67, 31); (68, 30); (69, 25); (70, 38);
   (71, 24); (72, 23); (73, 30); (74, 24); (75, 25); (76, 41); (77, 20);
   (78, 18); (79, 27); (80, 28); (81, 27); (82, 26); (83, 26); (84, 28);
   (85, 17); (86, 28); (87, 31); (88, 14); (89, 20); (90, 26); (91, 23);
   (92, 20); (93, 32); (94, 33); (95, 27); (96, 20); (97, 24); (98, 32);
   (99, 12); ...]
```

With a uniform distribution, a `size` of `1000`, and 10,000 samples, I'd expected to see 10 empty lists, 10 singleton lists, 10 lists with two elements, etcetera, but I see 70 empty lists, 58 singleton lists, 57 lists with two elements, and so on.

Am I missing something?

kurtschelfthout:
Oh very interesting. It isn't but it should be. I'm going to take a look at that.

kurtschelfthout:
Aaaah!

The documentation of `sample` is incorrect. It picks a size _up to_ the given size. So sample in this case naturally skews to lower sizes.

If you use:

```
let c =
    Gen.resize 10 (Gen.listOf (Gen.constant 0))

c
|> Gen.sample 10 1000
|> Seq.countBy List.length
|> Seq.sortBy fst
|> Seq.toArray
```

Then it's uniformly distributed.

We should fix `sample` to fix the size instead of picking one at random up to the given size.

ploeh:
Yes, using `Gen.resize` first makes the distribution uniform.

> We should fix sample to fix the size instead of picking one at random up to the given size.

While investigating this behaviour, I noticed that `Gen.sample` has that behaviour, but I thought it was in order to emulate the behaviour of exercising 'real' properties. Perhaps I'm reading too much into this, but doesn't the runner start with a size equal to `config.StartSize` and incrementally increase it to `config.EndSize`?

If I run a test using the default value of 100 test runs, doesn't it mean that I'm still not going to see a uniform distribution of list length over those 100 executions? Isn't it still going to be skewed towards smaller lists, including empty lists?

---

Anyway, how about a much simpler 'fix' that removes the `+1` from `listOf`?

```F#
let listOf gn =
    sized <| fun n ->
        gen {   let! k = choose (0,n)
                return! listOfLength k gn }
```

What is the underlying reason for the `+1` part?

kurtschelfthout:
You are right that the runner does that. It first picks a random size between 0 and the current size of the run, while all generators already interpret size as an upper bound, see `listOf`. So it's skewed downward twice if you will - once in the runner/sample and once by the generators themselves. I think the runner should at iteration n use size f(n) which is non-random.

To put it another way, the idea is that as more tests pass, the size gradually increases because we'll get more complex examples that way.

Right now it does on iteration i:
* Pick a random number between 0..i - that is the size for this run. Let's say this is j.
* Each generator that takes size into account does this by first picking a number between 0..j

I think if that change is made there is little reason left for the `+1`.

ploeh:
All of that makes sense :+1: except that I don't understand your last sentence:

> I think if that change is made there is little reason left for the `+1`.

Which change do you have in mind?

kurtschelfthout:
Change the runner to not pick a random number between 0 and size, but just use the current size.

ploeh:
Where does it do that?

kurtschelfthout:
Via `Gen.eval`. 

I think this change is perhaps too big to do in anything but a major version change. Although it's not strictly changing any API, it does change the behavior fairly significantly. Then again, bug fixes also change behavior of course.

Thoughts?

moodmosaic:
That sounds weird to me, since the entry point of the runtime size *is* the `Gen.eval` function.

---

Actually, I thought `eval` run generators with a *fixed* size, and if you want another size then you could use `resize` but that isn't the case here...

kurtschelfthout:
https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Gen.fs#L148

You can still use `resize`, as that effectively ignores whatever size was passed in, but that's not the point here. I agree it's weird that `eval` chooses a size between 0 and `n`. That's what I want to change :)

ploeh:
> Thoughts?

I never rely on distributions when writing properties, but perhaps other people do... I'm going to defer that decision to you, since you have a deeper insight into these aspects than I do.

In any case, I'm closing the PR, because whatever happens, it isn't going to be this implementation :smiley:

moodmosaic:
> I agree it's weird that `eval` chooses a size between 0 and `n`. That's what I want to change :)

Size could be initially set to some reasonable default. For example, [QuickCheck sets the initial size to `30`](https://github.com/nick8325/quickcheck/blob/6c360c1b594b0243e51bfcd4e2428882570e7a51/Test/QuickCheck/Gen.hs#L98):

```haskell
-- | Run a generator. The size passed to the generator is always 30;
-- if you want another size then you should explicitly use 'resize'.
generate :: Gen a -> IO a
generate (MkGen g) =
  do r <- newQCGen
     return (g r 30)
```

kurtschelfthout:
See https://github.com/fscheck/FsCheck/pull/264

PR #243, opened 2016-05-12
Made the role of the size argument more explicit

ploeh:
in documentation examples.

This pull request is a response to the comments about #238.


moodmosaic:
Looks good to me :smiley: :+1:

PR #244, opened 2016-05-12
Added Gen.filter as an alias of Gen.suchThat.

ploeh:
Many common monads have a 'filter' function. Examples include `List.filter`, `Map.filter`, `Set.filter`, `Option.filter`, and even `Arb.filter`. For the beginner, it's confusing that there's no `Gen.filter`; case in point: http://stackoverflow.com/q/37189368/126014

Even after I was able to localise `Gen.suchThat`, for a long time I was apprehensive of using it because I felt that I didn't sufficiently understand how it was different from a 'normal' filter function.

Since `Gen.filter` is simply a synonym for `Gen.suchThat`, I didn't add any tests. Likewise, there are no tests of `Gen.where`. I'll be happy to add a property for it, it this is wanted.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -376,6 +376,13 @@ module Gen =
     //[category: Creating generators from generators]
     [<CompiledName("Where");EditorBrowsable(EditorBrowsableState.Never)>]
     let where predicate generator = suchThat predicate generator
+    
+    ///Generates a value that satisfies a predicate. Contrary to suchThatOption, this function keeps re-trying
+    ///by increasing the size of the original generator ad infinitum.  Make sure there is a high probability that 
+    ///the predicate is satisfied.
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Should we deprecate `suchThat`? Any suggestions for another name for `suchThatOption`?

mausch:
+1 to deprecate `suchThat`.
`suchThatOption` could be renamed to `tryFilter` / `tryFilterOnce` ?

ploeh:
`Gen.tryFilter`?

ploeh:
Great minds think alike :smiley:

mausch:
Either that or we screw up alike :stuck_out_tongue_winking_eye:

kurtschelfthout:
I like tryFilter and tryWhere. Thanks.

ploeh:
Submitted as #251

PR #247, opened 2016-05-14
Removed Seq.toList from generator examples,

ploeh:
because I realized that `Gen.sample` already returns lists. When I originally wrote the samples, I thought that `Gen.sample` returned seqs. For documentation purposes, it's important to return lists or arrays instead of seqs, because seqs don't always render in a way that's useful for the reader.

(BTW, please accept my apologies for all these very fine-grained pull requests. I'm actually attempting to write some _more_ documentation examples, but I keep discovering these small things that I'd like to correct, and I still believe that it's better to keep each pull request small and focused.)


kurtschelfthout:
Have no problem with the small PRs, whatever is convenient for you.

PR #248, opened 2016-05-14
Add generator for System.Net.IPAddress

ErikSchierboom:
Add a generator for [`System.Net.Mail.IPAddress`](https://msdn.microsoft.com/en-us/library/system.net.ipaddress(v=vs.110).aspx). As this class is defined in `System.dll`, there are no additional dependencies.

The only thing noteworthy is that this class does not exist in any of the PCL targets, hence the `#if` clause.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -11,6 +11,7 @@
 namespace FsCheck
 
 open System
+open System.Net
 
 ///Represents an int >= 0
 type NonNegativeInt = NonNegativeInt of int with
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
By the way, I was not sure how to shrink an IPAddress instance, which I why I didn't include a shrinker. Is that something you'd like added?

kurtschelfthout:
Thanks!

Shrinker would be good, perhaps just try special values like 127.0.0.1? Or try making everything as close to 0? Often even a very simple shrink is useful as it tells you at least that a value can be made smaller or not.

ErikSchierboom:
@kurtschelfthout I've modified the PR to also add a shrinker. But I do is just using the default shrink strategy for `byte[]`'s and then filtering the result to make sure the length stays the same. Is that a sane approach?

Furthermore, I refactored the generator as well to make it simpler.

kurtschelfthout:
Nice! Can you also add a test for the shrinker?

If you're out of time happy to merge as is and add later, let me know.

ErikSchierboom:
Totally forgot about adding a test for the shrinker, sorry. I have just updated the PR with the added shrink test.

kurtschelfthout:
Thank you!

ErikSchierboom:
You're welcome!

PR #253, opened 2016-05-14
Provide examples for Gen.choose

ploeh:
Another addition related to #218

### DIFF
--- a/docs/content/TestData.fsx
+++ b/docs/content/TestData.fsx
@@ -249,17 +249,17 @@ values drawn from a collection of possible values. The collection is inclusive,
 which means that both the first and last element, as well as all elements
 between, can be drawn.
 
-In the following example, the numbers from 0 to 9 define the collection of
+In the following example, a list of arbitrary integers define the collection of
 possible values. The result is a generator that creates `int` values guaranteed
-to be between 0 and 9 (both included). Since `Gen.elements` doesn't rely on the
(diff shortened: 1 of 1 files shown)


docs/content/TestData.fsx
```diff
@@ -208,24 +208,58 @@ all singular elements returned is the same tuple.*)
...
+#### Choose
+
+You can use the `Gen.choose` function to create a generator of singular integer
+values between a minimum and maximum value, both inclusive:*)
+
+(***define-output:ChooseBetweenZeroAndNineExample***)
+Gen.choose (0, 9) |> Gen.sample 0 10
+
+(**This example generates a single integer value between 0 and 9. Since
+`Gen.constant` doesn't rely on the `size` argument, it's `0` in this example,
```
kurtschelfthout: Typo here - `Gen.choose`, not `Gen.constant`. Do we need to repeat this same line every time?

ploeh: Well spotted :smile: Pull request updated.

> Do we need to repeat this same line every time?

We don't need to, but my approach so far has been to write each section to be as self-contained as possible. The intent is that the documentation can act as a reference more than a narrative.

It does lead to repetition, but I think it's safer than assuming that the reader has read everything from the top.

I'll try to vary my prose a bit in the future :flushed:

PR #254, opened 2016-05-14
Added examples of list generators.

ploeh:


ploeh:
Another addition related to #218

Issue #200, opened 2016-03-03
Support generating MailAddress instances


ErikSchierboom:
It would be great if FsCheck would be able to generate `MailAddress` instances. I could work on this, but first I'd like to know if this is something that you'd support.

mausch:
MailAddress is in System.dll so it would represent no additional dependencies for FsCheck so I don't see why not.

ErikSchierboom:
Okay, great! I'll have a go at it then.

ErikSchierboom:
I'm working on adding this feature, but I have one important question: email addresses have quite a [complex format](https://en.wikipedia.org/wiki/Email_address). Should I create a generator that can randomly generate every type of valid email address? If so, the generator will probably be quite complex.

If not, should I only generate "simple" email addresses, or should I use a list of pre-defined email addresses which contains all edge cases? That list could look something like this:

- prettyandsimple@example.com
- very.common@example.com
- disposable.style.email.with+symbol@example.com
- other.email-with-dash@example.com
- x@example.com (one-letter local part)
- "much.more unusual"@example.com
- "very.unusual.@.unusual.com"@example.com
- "very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com
- example-indeed@strange-example.com
- admin@mailserver1 (local domain name with no TLD)
- #!$%&'*+-/=?^_`{}|~@example.org
- "()<>[]:,;@\\\"!#$%&'*+-/=?^_`{}| ~.a"@example.org
- " "@example.org (space between the quotes)
- example@localhost (sent from localhost)
- example@s.solutions (see the List of Internet top-level domains)
- user@com
- user@localserver
- user@[IPv6:2001:db8::1]

What would you consider to be the best approach?

kurtschelfthout:
The "usual" approach is to have a (relatively) general generator, augmented with special cases that have a high probability of being generated. Have a look at Arb.Default.Float as an example: it generates floats, but with a high chance of hitting epsilon, min value etc. 

Does that make sense?

ErikSchierboom:
Yes it does. Any preferences to how the general generator should work? More specifically, should I generate random strings (of the correct length and using the correct characters of course), or use a pre-defined set of values? People don't expect an email address to look like "a2#$*/e@example.com" or something like that, although that is of what FsCheck can help you with, testing every valid email address.

kurtschelfthout:
[...]

PR #255, opened 2016-05-15
Add support for generating MailAddress (#200)

ErikSchierboom:
This PR is the first attempt to add support for generating `System.Net.Mail.MailAddress` instances.

As noted in #200, the list op top-level domains is quite large. Maybe you want this changed in the code? I used [RFC 3696](https://tools.ietf.org/html/rfc3696#page-3) to guide my implementation. Some oddities were found where the .NET framework would not accept a mail address that was valid according to the specification, so I left those out.

At the moment, I haven't added a shrinker yet. The most obvious suggestion would be to make the email address smaller, but how would that work with top-level domains?


kurtschelfthout:
Interesting! I thought this would be a rather easy one but I was totally naive!

I have a few remarks.
 1. The test on Travis fails. I imagine that mono's MailAddress is even more restrictive than .NET's...
 2. I am ok with having an exhaustive list, but would like to keep it out of the code as it's pretty long. Embedded resource perhaps? http://stackoverflow.com/questions/2689446/how-to-manage-resources-in-an-f-project

By the way I know there is already a list of cultures in there, I intend to move those too, but that is not your problem...

As for top-level domains I would suggest just shrinking to "obvious" well known ones like com, net, etc.

ploeh:
@kurtschelfthout What's the problem with having a long list in code? How does moving it to an embedded resource address that?

kurtschelfthout:
Readability of that particular code file. Basically, I don't like scrolling. If F# would have folding support, I probably wouldn't mind so much.

ErikSchierboom:
Maybe we could move those long lists to another .fs file instead of a resource?

I'll look into the Travis build failing on my Mac.

kurtschelfthout:
@ErikSchierboom sure, that's fine too.

ErikSchierboom:
Small question about my shrinker. How would I best go about that? My initial attempt was as follows:

```fsharp
let isValidQuotedUser (user: string) = user.Length > 1 && user.StartsWith("\"") && user.EndsWith("\"")
let isValidUnquotedUser (user: string) = 
    not (user.StartsWith("\"")) && 
    not (user.EndsWith("\"")) && 
    not (user.StartsWith(".")) && 
    not (user.EndsWith(".")) && 
    not (user.Contains(".."))

let isValidUser (user: string) = isValidQuotedUser user || isValidUnquotedUser user
let isValidHost = not << String.IsNullOrWhiteSpace

let shrinker (a:MailAddress) = 
    seq {
        for name in seq { yield a.DisplayName; yield! shrink a.DisplayName } do
        for user in seq { yield a.User; yield! shrink a.User } |> Seq.filter isValidUser do
        for host in seq { yield a.Host; yield! shrink a.Host; yield! commonTopLevelDomains } |> Seq.filter isValidHost do   
            if String.IsNullOrWhiteSpace name then                     
                yield MailAddress (sprintf "%s@%s" user host)                              
            else 
                yield MailAddress (sprintf "%s <%s@%s>" name user host)                              
    } |> Seq.skip 1   
```

However, that generates lots of shrinks due to the cartesian product being used. What would be the best strategy here?

kurtschelfthout:
The best way to go about it is to try and shrink each part separately, in some order. So first shrink the `DisplayName`, leaving the rest untouched in a separate loop. Then shrink the user, again leaving the rest untouched. Then shrink host, leaving the rest untouched. Shrinking happens recursively, i.e. when FsCheck finds a smaller value, it will again try to shrink that, so you don't need to generate the cartesian product.

Also no need to yield the original `MailAddress` and then `skip 1`, just don't yield it...

ErikSchierboom:
I came up with the following shrinker:

```fsharp
let shrinker (a:MailAddress) = 
    seq {
        yield! Arb.shrink a.DisplayName |> Seq.map (fun name -> createMailAddress name a.User a.Host)
        yield! Arb.shrink a.User |> Seq.filter isValidUser |> Seq.map (fun user -> createMailAddress a.DisplayName user a.Host)
        yield! Arb.shrink a.Host |> Seq.filter isValidHost |> Seq.map (fun host -> createMailAddress a.DisplayName a.User host)
        yield! commonTopLevelDomains |> Seq.filter (fun host -> host <> a.Host) |> Seq.map (fun host -> createMailAddress a.DisplayName a.User host)             
    } 
```

Everything works, but the shrinker can take a bit long. I think this is due to the fact that too many shrink values are generated. For example,

```
let addresses = Arb.generate<MailAddress> |> Gen.sample 10 10
let shrinks = addresses |> Seq.collect shrink |> Seq.length
```

The number of shrinks generated for 10 samples hovers around 2500. I think this is much too much.   Any tips?

kurtschelfthout:
I am guessing this is because the string shrinker generates many values - if I remember correctly it converts the string to a char array and then shrinks that. That means removing each character in turn, then trying to shrink each character individually. 

Perhaps for the names, domains etc it would make sense to do something else - say trying to split the string in half, i.e. shrink to the first half, the second half, and maybe not even try to shrink the characters?

ErikSchierboom:
That seems like a sensible approach, I'll give that a try. Is there any functionality already in FsCheck that I can use to select a random index in the string on which to split? In other words, should I just use the `Random` class from the .NET framework, or should I be using the `Random` module from FsCheck? There is the `Gen.choose` function that allows a random integer within a range to be selected, however, that returns a Generator, and I want to return actual values. I could use `Gen.sample` for that, but that expects a size. I'm not sure how to do this. 

Similarly, suppose I want to change the host name randomly to one of the values in a pre-defined list. How would I do that best? There is a `Gen.elements` function that seems suitable, but that also returns a Generator, not an actual value.

kurtschelfthout:
I think there is some confusion here. This is for the shrinker so it should be fully deterministic, i.e. dont use anything in `Gen` or `Random`. So my idea is to split the string in the middle, and shrink to the first and second half, like this:

    let shrink str = seq { yield str.SubString(0,str.Length/2); yield str.SubString(str.Length/2,str.Length
1)

which basically results in a binary chop for each of the strings. Note there are a few edge cases to deal with in the above quick and dirty code....

ErikSchierboom:
@kurtschelfthout Thanks, I'll look into it.

ErikSchierboom:
@kurtschelfthout I finally managed to get things working :)

Issue #252, opened 2016-05-14
Experimental.StateMachine Behavior on unexpected input


MarcoMaar:
I was thinking of how we should handle the following two cases.
- No operation that satisfies the precondition could be generated.

Currently if the precondition is not satisfied the generated sequence to the current point is used. Alternately we could throw an exception and inform the user that the precondition couldn't be met. Otherwise the user may not realize that his model is faulty.
The downside would be that if we are at a terminal state the user would have to create a StopCommand by himself... So I am not sure what we could do, but I think we should not conceal that the Preconditions can't be satisfied.
- In Next no Operation is generated at all

Currently if no operation is generated in next this exception is thrown:
`System.IndexOutOfRangeException : Index was outside the bounds of the array.`

Maybe we could check for this behavior in FsCheck and throw a custom exception that informs the user he forgot to generate an operation.

A case where this error can occur is for example the following:
You use weights to generate commands. Especially if the weights are calculated in a not straightforward way it could happen that all weights are 0 without noticing it.

MarcoMaar:
And I found a third thing.

- Postcondition not satisfied.

Currently if `Machine.Check` fails the run is still continued until the end. I think it would be better to abort the run and list how far Check got in the sequence.

The reason why I would abort without completing the run is the following:
If our SUT enters an unexpected state we can no longer safely operate it. This can result in further errors. The problem is then that FsCheck might not report the first error but another one. Finding the root problem is then tough.

Oh .. We could also shorten the sequence for the Shrinker if we abort early.

kurtschelfthout:
Agree with the second and third points. 

First point too, I think it should generate an arguments exhausted message at some point - same as what you get if you have a strong filter and FsCheck can't find enough examples to satisfy the filter. 

Third point is the result of using .&. to recursively combine each step; that's been on my list to change for a while.

kurtschelfthout:
Should now be addressed.

PR #257, opened 2016-05-15
Behavior on unexpected input

MarcoMaar:
PR for #252 

- [ ] No operation that satisfies the precondition could be generated.
- [ ] In Next no Operation is generated at all.
- [ ] Postcondition not satisfied.

I started with the second point. I realized that the problem can be tackled directly at the generators of the Gen module

Question is: should I add exception handling for the generators that might not work correctly if invalid input is specified? Seems like a big change if I do that. What do you think?

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -202,11 +202,14 @@ module Gen =
 
     ///Build a generator that generates a value from one of the generators in the given non-empty seq, with
     ///given probabilities. The sum of the probabilities must be larger than zero.
+    ///<exception cref="System.ArgumentException">Thrown when the sum of the propabilites is smaller or equal 0.</exception>
     //[category: Creating generators from generators]
     [<CompiledName("Frequency")>]
     let frequency xs =
         let xs = Seq.toArray xs
         let tot = Array.sumBy fst xs
+        let throwIfInvalid = 
+            if tot <= 0 then invalidArg "xs" "Frequency was called with a sum of propabilites smaller or equal 0. No elements can be generated." 
         let rec pick i n =
             let k,x = xs.[i]
             if n<=k then x else pick (i+1) (n-k)
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -202,7 +202,7 @@ module Gen =
 
     ///Build a generator that generates a value from one of the generators in the given non-empty seq, with
     ///given probabilities. The sum of the probabilities must be larger than zero.
-    ///<exception cref="System.ArgumentException">Thrown if the sum of the probabilites is less than or equal to 0.</exception>
+    ///<exception cref="System.ArgumentException">Thrown if the sum of the probabilities is less than or equal to 0.</exception>
     //[category: Creating generators from generators]
     [<CompiledName("Frequency")>]
     let frequency xs =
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
I think having better error messages is overall an improvement. 

With respect to breaking changes, if the behaviour does not change, except perhaps for the type of exception thrown and a better error message, I think it is fine as the behaviour in that case was not documented/specified before, it threw anyway, and so we are not breaking anything. 

We should also consider for each case whether it's worth actually propagating the error in another way. For example in this case perhaps it is better to go with "arguments exhausted" rather than failing (probably not but it's worth considering).

I would encourage you to open a separate PR for each of your points, as well as for adding argument validation to functions in `Gen`, as otherwise things will get confusing very quickly.

src/FsCheck/Gen.fs
```diff
@@ -202,11 +202,14 @@ module Gen =
 
     ///Build a generator that generates a value from one of the generators in the given non-empty seq, with
     ///given probabilities. The sum of the probabilities must be larger than zero.
+    ///<exception cref="System.ArgumentException">Thrown when the sum of the propabilites is smaller or equal 0.</exception>
```
kurtschelfthout: Some corrections: "probability", smaller _than_

src/FsCheck/Gen.fs
```diff
@@ -202,11 +202,14 @@ module Gen =
...
     //[category: Creating generators from generators]
     [<CompiledName("Frequency")>]
     let frequency xs =
         let xs = Seq.toArray xs
         let tot = Array.sumBy fst xs
+        let throwIfInvalid = 
+            if tot <= 0 then invalidArg "xs" "Frequency was called with a sum of propabilites smaller or equal 0. No elements can be generated."
```
kurtschelfthout: See above for corrections in text, also prefer just writing the if, i.e. omit the `let throwIfInvalid`.

tests/FsCheck.Test/Gen.fs
```diff
@@ -64,6 +64,12 @@ module Gen =
                 |> Gen.frequency
                 |> sample 100
                 |> List.forall (isIn generatedValues)))
+
+    [<Fact>]
+    let ``frequency should realize if no element can be generated``() =
```
kurtschelfthout: Suggest: frequency should _throw_ if no element can be generated

MarcoMaar:
> We should also consider for each case whether it's worth actually propagating the error in another way. For example in this case perhaps it is better to go with "arguments exhausted" rather than failing (probably not but it's worth considering).

I guess it could be nice to propagate some invalid input via the test results. In this particular case the generator throws an exception upon building. This means this can happen outside of a test run.
```csharp
Func<int, bool> prop = i => true;
var arb = Arb.From(Gen.Frequency(new Tuple<int, Gen<int>>(0, Gen.Constant(1))));
Prop.ForAll(arb, prop).QuickCheck();
```
The code will fail in line 2 already, before using the generator to generate values. So I can't think of another way than throwing an exception. 

> 
I would encourage you to open a separate PR for each of your points, as well as for adding argument validation to functions in Gen, as otherwise things will get confusing very quickly.

Should I delete this PR and create a new one or should I start doing this with the next small change?

kurtschelfthout:
Just start with the next one I think. Thanks.

kurtschelfthout:
Re: throwing the exception, yes I think it's ok in this case. That's what it was doing before anyway.

PR #258, opened 2016-05-17
Fix Mono compilation errors

ErikSchierboom:
I was opening the `FsCheck-mono.sln` solution file in Xamarin Studio on my Mac, and found that the code did not compile. Both errors had the same cause: the deprecated `Seq.nth` was used instead of `Seq.item`:

image

This PR fixes this.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -166,7 +166,7 @@ module Gen =
     //[category: Creating generators]
     [<CompiledName("Elements")>]
     let elements xs = 
-        choose (0, (Seq.length xs)-1) |> map (flip Seq.nth xs)
+        choose (0, (Seq.length xs)-1) |> map (flip Seq.item xs)
 
     [<CompiledName("GrowingElements")>]
     ///Build a generator that takes a non-empty sequence and randomly generates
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Hm, that doesn't sound right. The AppVeyor (windows) build fails now. I think for some reason your build is using a later version of FSharp.Core?

ErikSchierboom:
Hmm, that is odd. I just opened the Mono solution in the latest Xamarin Studio. Could it be that the Mono solution does not explicitly specify the FSharp.Core version or something like that?

kurtschelfthout:
I don't quite understand myself :) Also the travis solution does _not_ fail either before or after your change so I'm not sure what's going on!

ErikSchierboom:
Odd. I'll see if I can find anything about this problem.

ErikSchierboom:
I looked into it, and it appears to be a Xamarin Studio bug. The `FsCheck.fsproj` file lists the following explicit FSharp.Core version:

```xml
<TargetFSharpCoreVersion>4.3.1.0</TargetFSharpCoreVersion> 
```

However, if I right-click on the FSharp.Core reference in the Mono solution, it lists that it uses version 4.4.0.0:

image

I think this is a Xamarin Studio bug, so I'll file a bug there.

ErikSchierboom:
This is the [bug I filed](https://bugzilla.xamarin.com/show_bug.cgi?id=41166) by the way.

kurtschelfthout:
Thanks!

PR #262, opened 2016-05-19
Gen.Frequency Xml Comments

MarcoMaar:
PR related to #252

Changed the xml comments of frequency to use xml tags and added them to the overloaded methods.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -200,9 +200,12 @@ module Gen =
     let oneOfArr ([<ParamArrayAttribute>]  generators : array<Gen<_>>) = 
         generators |> oneof
 
-    ///Build a generator that generates a value from one of the generators in the given non-empty seq, with
-    ///given probabilities. The sum of the probabilities must be larger than zero.
-    ///<exception cref="System.ArgumentException">Thrown if the sum of the probabilities is less than or equal to 0.</exception>
+    /// <summary>
+    /// Build a generator that generates a value from one of the generators in the given non-empty seq, with
+    /// given probabilities. The sum of the probabilities must be larger than zero.
+    /// </summary>
+    /// <param name="xs">Sequence of tuples where each tuple contains a weight and a generator.</param>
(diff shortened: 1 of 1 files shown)


MarcoMaar:
maybe we should change the parameter in
https://github.com/fscheck/FsCheck/pull/262/files#diff-cf424f2bdbcba155096e701a72328711R211
to weightedValue to fit the other functions
https://github.com/fscheck/FsCheck/pull/262/files#diff-cf424f2bdbcba155096e701a72328711R233

kurtschelfthout:
No, don't think that's worth changing. 2-tuples are fine.

MarcoMaar:
I just meant the name of the parameter. It is "xs" in the first function and "weightedValues" in the other 3. But, yeah isn't a big deal I guess.

kurtschelfthout:
Oh, I see, I misunderstood. Yes, though even `weightedValues` is probably a bad name.Because they're generators, not values. Maybe `frequencies` would be better...

MarcoMaar:
I was thinking of changing the name to `weightedGenerators`, since I also used it in the comments. Then I started thinking about the names and looked up some definitions...

`frequencies` and `weightedGenerators` are not entirely correct I think. `frequency` is the number of times an event occurs. Therefore there should not be a generator within it. `weightedValue` is the product between the value and the weight. `WeightAndGenerators` would therefore be closer to correct since no calculation was performed.

There is also the record type `WeightAndValue<'a>' with `Value : 'a`. I think it would be more clear to change it to `WeightAndGenerator<'a>' with `Generator: 'Gen<a>` but this would be a breaking change.

I would leave `WeightAndValue<'a>' as it is, since it is breaking. I would change the parameters to `weightedGenerators`. Even though it is not 100% correct it transmits what the parameter will be used for.

Sorry, if you feel like I am wasting our time with unimportant details let me know and I will not focus on them that much.

kurtschelfthout:
Well, perhaps we're over-thinking it a bit :) I don't think it's worth a breaking change at this point, but we could change it on the 3.0 branch. But  to be honest I think I introduced that type before C# had tuples, I think by now we should just use tuples as arguments for C#.

PR #264, opened 2016-05-19
Gen.eval uses size, instead of up to size.

kurtschelfthout:
This is discussed somewhat further in PR 241. The underlying issue is
that before this change, there was a bias to generate values of low
size. This is because in `Gen.eval`, which is also used by the runner directly
to evaluate a generator, a random size between 0 and the current test
size (which is set in configuration) is picked. This size is then passed
on to the generator. But, almost all generators that take size into account
themselves also pick a random size between 0 and the given size. So
overall, the size was skewed downward. This change addresses that by
changing eval to simply use the size as given. By default, FsCheck will
increase the size linearly from 1 to 100 over the course of the number
of tests is runs (also 100 by default).

I have also changed the `listOf` generator to now pick a size between
`0` and `n`, instead of `n+1` as that was a hack to decrease the chance
of generating quite so many empty lists - this should no longer be
necessary.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -141,20 +141,19 @@ module Gen =
     [<CompiledName("Resize")>]
     let resize newSize (Gen m) = Gen (fun _ r -> m newSize r)
 
-    ///Generates a value with maximum size n.
+    ///Generates a value of the give size with the given seed.
     //[category: Generating test values]
     [<CompiledName("Eval")>]
-    let eval n rnd (Gen m) = 
-        let size,rnd' = range (0,n) rnd
-        m size rnd'
+    let eval size seed (Gen m) =
+        m size seed
 
     ///Generates n values of the given size.
     //[category: Generating test values]
     [<CompiledName("Sample")>]
-    let sample size n gn  = 
+    let sample size n generator  = 
         let rec sample i seed samples =
             if i = 0 then samples
-            else sample (i-1) (Random.stdSplit seed |> snd) (eval size seed gn :: samples)
+            else sample (i-1) (Random.stdSplit seed |> snd) (eval size seed generator :: samples)
         sample n (Random.newSeed()) []
 
     ///Generates an integer between l and h, inclusive.
@@ -422,7 +421,7 @@ module Gen =
     [<CompiledName("ListOf")>]
     let listOf gn =
         sized <| fun n ->
-            gen {   let! k = choose (0,n+1) //decrease chance of empty list
+            gen {   let! k = choose (0,n)
                     return! listOfLength k gn }
 
     /// Generates a non-empty list of random length. The maximum length 
--- a/src/FsCheck/Script.fsx
+++ b/src/FsCheck/Script.fsx
@@ -7,6 +7,16 @@ open FsCheck
 open FsCheck.Experimental
 open System.Collections.Generic
 
+let l = Gen.listOf (Gen.constant 1)
+l 
+|> Gen.sample 10 1000
+|> Seq.map Seq.length 
+|> Seq.groupBy id 
+|> Seq.map (fun (l,gr) -> (l, Seq.length gr))
+|> Seq.sortBy fst
+|> Seq.toArray
+
+
 type Pid = int
 type Name = string
 type ProcessRegistry() =


src/FsCheck/Script.fsx
```diff
@@ -7,6 +7,16 @@ open FsCheck
...
 open System.Collections.Generic
 
+let l = Gen.listOf (Gen.constant 1)
+l 
+|> Gen.sample 10 1000
+|> Seq.map Seq.length 
+|> Seq.groupBy id 
+|> Seq.map (fun (l,gr) -> (l, Seq.length gr))
+|> Seq.sortBy fst
+|> Seq.toArray
```
kurtschelfthout: This script here is just to show that after this change, lengths are uniformly distributed.

kurtschelfthout:
Few points I'd like to discuss.
1. Is this a breaking change? Clearly the API doesn't change. But, the likelihood of bigger sized values is markedly increased. But but, some of these values were already generated before. So, my sense is we can get away with a minor version increase (clearly a patch version is not appropriate)
2. A side-effect is that the tests take longer. Perhaps by default we should decrease the maximum test size to 50 from 100, i.e. FsCheck increases the size linearly from 1 to 50. This means by default it will generate smaller values, increasing performance but removing a bunch of values of size > 50 that were previously generated. Paradoxically this makes me feel that it makes the change "more breaking",

Thoughts?

src/FsCheck/Gen.fs
```diff
@@ -141,20 +141,19 @@ module Gen =
...
     let resize newSize (Gen m) = Gen (fun _ r -> m newSize r)
 
-    ///Generates a value with maximum size n.
+    ///Generates a value of the give size with the given seed.
     //[category: Generating test values]
     [<CompiledName("Eval")>]
-    let eval n rnd (Gen m) = 
-        let size,rnd' = range (0,n) rnd
-        m size rnd'
+    let eval size seed (Gen m) =
```
moodmosaic: Nice. The `eval` no longer takes a `rnd` :+1: It's much easier to use now outside of the test runner (e.g. when defining a custom `sample` function).

moodmosaic:
>1. Is this a breaking change? Clearly the API doesn't change.

No, the API *does* change since `Gen.eval` has changed, which is a public function.

>2. [..] Perhaps by default we should decrease the maximum test size to 50 from 100

This looks like a breaking change, as I could imagine an FsCheck user upgrading to the latest version and saying *"WTF?! Why my properties are executed half of the times?!"*.

---

All in all, this PR looks great, but I'd only push in on some `v3` branch, considering also #236 as well.

src/FsCheck/Gen.fs
kurtschelfthout: No, the signature is the same as before. `int -> StdGen -> Gen<'T> -> 'T`. Just the name of the argument changed.

kurtschelfthout:
@moodmosaic 

> No, the API does change since Gen.eval has changed, which is a public function.

The signature has not changed. The behavior has changed, clearly, but the "policy" decision we have to make here is: is changing the distribution of values a major "breaking" change or a minor change?

I have made both calls in the past. Am thinking of making a policy more explicit going forward.

> This looks like a breaking change, as I could imagine an FsCheck user upgrading to the latest 
> version and saying "WTF?! Why my properties are executed half of the times?!".

It's not half the times, same number of times except the size would increase from 1 to 50 instead of 1 to 100. So size per test would be 1,1,2,2,3,3,...50,50 instead of 1,2,3,4,5,6,7,...100

src/FsCheck/Gen.fs
moodmosaic: Oh, yes indeed.

moodmosaic: But then, `seed` isn't the best name...

moodmosaic:
>is changing the distribution of values a major "breaking" change or a minor change?

I wouldn't take that as a breaking change; it's not *explicitly* stated anywhere (?) – unless I miss something.

>It's not half the times, same number of times except the size would increase from 1 to 50 instead of 1 to 100.

Fair enough - that shouldn't be a breaking change either. The runtime size  can be volatile and if a property relies on that then it's a smell; in such cases a cunstom gen/shrink pair is warranted.

---

>So, my sense is we can get away with a minor version increase

That makes sense now, IMHO.

ploeh:
This looks good, and I wouldn't consider any of it a breaking change.

AFAIR, FsCheck follows [Semantic Versioning 2.0](http://semver.org/spec/v2.0.0.html), and according to that definition, this isn't even a minor release, since a minor release is "when you add functionality in a backwards-compatible manner". There's no functionality added here.

Is it a breaking change that behaviour changes? Only, in my opinion, if that change in behaviour affects users.

Changing the distribution will most likely cause tests to run a bit slower, but they also now provide more coverage. It's a trade-off, but I don't think's it's a breaking change.

I can't think of a good scenario where someone would write a property that implicitly relies on the random distribution of generated values, or a property that relies on timing.

If someone else can provide a good example of such tests, I can see how this proposed change would be a breaking change.

kurtschelfthout:
Ok, that sounds reasonable. Perhaps I have been a bit too conservative in not allowing some changes.

PR #265, opened 2016-05-20
Corrected a small typo in the docs

moodmosaic:
I could be wrong, as I'm not a native English speaker. Though, as per [this](http://english.stackexchange.com/a/101333) thread, 'both' implies a plural.


kurtschelfthout:
I read this as short for "both the first element and the last element...can be drawn" so technically correct use of plural but probably superfluous usage of "both". Can just be omitted.

moodmosaic:
:+1: Then we can close this.

PR #269, opened 2016-05-26
Gave examples of values generated with Gen.filter.

ploeh:
This is another pull request related to #218.

### DIFF
--- a/docs/content/TestData.fsx
+++ b/docs/content/TestData.fsx
@@ -496,6 +496,34 @@ function that takes a four-element tuple.*)
 in order to produce a more readable output. The resulting generator has the
 type `Gen<string>`, but if you remove the `string` composition, the type would
 be `Gen<Version>.`
+
+#### Filter
+
+While you can use the above generators and combinators to define various custom
+rules for generating values, occasionally you have a requirement where the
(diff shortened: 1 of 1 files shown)


moodmosaic:
:+1:

kurtschelfthout:
This is great and I will merge as is, the following are thoughts for your consideration.

- Mention `Gen.where`. Despite `filter` being the canonical name from FP, I think `where` is the clearer name. With `filter` somehow it's less intuitive to me what to return, true or false, to remove an element...
- Mention in the last paragraph what the effect is if there is a lower chance of the filter being satisfied (slower running or non-terminating tests)
- which leads to mentioning `tryWhere` as a safer alternative.

ploeh:
I've created #270 to address some of these comments, although I chose to use the name `Gen.tryFilter` instead of `Gen.tryWhere`, since this section exclusively discusses `Gen.filter`.

It'd be a good idea to document that `Gen.where` is identical to `Gen.filter`, but I'm not sure these example sections are the right place to do that. Wouldn't it be more appropriate to put that information in the API documentation?

We have to be careful not to put _everything_ in these example sections, as they can also grow so big that they become 'too much' for the reader... I think...

PR #271, opened 2016-05-27
ExceptionHandling: No Operation satisfies Precondition

MarcoMaar:
PR related to #252

now throws an exception if no command can be generated after a certain
number of retries.

### DIFF
--- a/src/FsCheck/StateMachine.fs
+++ b/src/FsCheck/StateMachine.fs
@@ -246,14 +246,25 @@ module StateMachine =
 
     [<CompiledName("Generate")>]
     let generate (spec:Machine<'Actual,'Model>) = 
-        let rec genCommandsS state size =
+        let retryNumber = 100
+        let rec genCommandsS state size retries =
             gen {
                 if size > 0 then
                     let nextOperation = spec.Next state
                     let! command = nextOperation |> Gen.tryWhere (fun operation -> operation.Pre state)
-                    if Option.isNone command || command.Value.GetType() = typeof<StopOperation<'Actual,'Model>> then return [state],[]
+                    if Option.isNone command then
+                        if retries > 0 then
+                            let! states, commands = genCommandsS state size (retries-1) //retry command generation
+                            return states, commands
+                        else
+                            let message = sprintf "It was not possible to generate a command after %d retries. \
+                                                  Maybe no precondition could be satisfied." retryNumber
+                            invalidArg "command" message
+                            return [],[] //is neeeded by the type checker, even though exception is thrown beforehand
+                    elif command.Value.GetType() = typeof<StopOperation<'Actual,'Model>> then 
+                        return [state],[]
                     else
-                        let! states, commands = genCommandsS (command.Value.Run state) (size-1)
+                        let! states, commands = genCommandsS (command.Value.Run state) (size-1) retryNumber
                         return state :: states, command.Value :: commands
                 else
                     return [state],[]
@@ -261,7 +272,7 @@ module StateMachine =
         gen { let! setup = spec.Setup |> Arb.toGen
               let initialModel = setup.Model()
               let maxNum = spec.MaxNumberOfCommands
-              let! models,operations = Gen.sized (fun s -> let size = if maxNum < 0 then s else maxNum in genCommandsS initialModel size)
+              let! models,operations = Gen.sized (fun s -> let size = if maxNum < 0 then s else maxNum in genCommandsS initialModel size retryNumber)
               return { Setup = initialModel, setup
                        Operations = List.zip operations (List.tail models) //first state is actually the initial state; so drop it.
                        TearDown = spec.TearDown }
--- a/src/FsCheck/StateMachine.fs
+++ b/src/FsCheck/StateMachine.fs
@@ -118,10 +118,12 @@ type Machine<'Actual,'Model>(maxNumberOfCommands:int) =
     abstract Next : 'Model -> Gen<Operation<'Actual,'Model>>
 
 [<StructuredFormatDisplayAttribute("{StructuredToString}")>]
-type MachineRun<'Actual, 'Model> =
-    { Setup : 'Model * Setup<'Actual,'Model> 
-      Operations : list<Operation<'Actual,'Model> * 'Model>
-      TearDown : TearDown<'Actual> } with
+type MachineRun<'Actual, 'Model>(setup, operations, tearDown) =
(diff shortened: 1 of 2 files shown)


MarcoMaar:
Retries a certain number of times (100) if the precondition for the command in next was not satisfied.

Here are some points to think of:

- Now terminal states have to be made explicit by either staying in a terminal state or calling a stop command in them. Before the sequence would just end. Se adjusted test case of symbolic machine for how this affects the machines.
https://github.com/MarcoMaar/FsCheck/blob/3979fb4f25338bcc0562e1a9767950ad44fa46e0/tests/FsCheck.Test/StateMachine.fs#L252

- Maybe make the settings number of retries adjustable.
https://github.com/MarcoMaar/FsCheck/blob/3979fb4f25338bcc0562e1a9767950ad44fa46e0/src/FsCheck/StateMachine.fs#L249

- This line seems ugly to me:
https://github.com/MarcoMaar/FsCheck/blob/3979fb4f25338bcc0562e1a9767950ad44fa46e0/src/FsCheck/StateMachine.fs#L263

src/FsCheck/StateMachine.fs
```diff
@@ -246,22 +246,33 @@ module StateMachine =
...
-        let rec genCommandsS state size =
+        let retryNumber = 100
+        let rec genCommandsS state size retries =
             gen {
                 if size > 0 then
                     let nextOperation = spec.Next state
                     let! command = nextOperation |> Gen.tryWhere (fun operation -> operation.Pre state)
-                    if Option.isNone command || command.Value.GetType() = typeof<StopOperation<'Actual,'Model>> then return [state],[]
+                    if Option.isNone command then
+                        if retries > 0 then
```
kurtschelfthout: I don't quite understand why we need the retries. `tryWhere` already has retries built-in.

kurtschelfthout:
I'm wondering if this is the right behaviour. An exception seems a bit excessive. Perhaps only throw an exception if from the initial state no next states can be found?

MarcoMaar:
I am having problems fully understanding the computation expression. So please correct me if I am saying something wrong or tell me if some of my approaches are way off.

> I don't quite understand why we need the retries. tryWhere already has retries built-in. 

tryWhere is related to the size of the generator. Since the size will be small for the first test runs tryWhere will not be able to find a command that satisfies the precondition if the likelihood of a "good" command is way lower than the likelihood of a "bad" command.
E.g.: https://github.com/fscheck/FsCheck/pull/271/files#diff-752b316be1f97e13cd2128ce5b04c1a1R157
Maybe the retries are stupid and there isa different way but I think Next should have a good chance in finding a usable command. 

What I am currently doing in production is only generating commands which satisfy their precondition since I can't find a solution to make it likely that tryWhere finds a command. But that seems like a workaround to me.

> I'm wondering if this is the right behaviour. An exception seems a bit excessive. Perhaps only throw an exception if from the initial state no next states can be found?

I agree that an exception might be a bit extreme. However, what I don't want is to pretend that the test run succeeded without any issues. It aborted premature and we are not propagating this to the user. I guess a warning would be nice. But I don't know an elegant way to do that.

kurtschelfthout:
`tryWhere` already retries to generate a value multiple times - sure it tries to increase the size but that's not really my point. If the precondition has such a low chance of being satisfied that it can't find an operation, the right way to respond is to return from the `Operation` a set of next operations with preconditions that have a good chance of being satisfied. I don't think that is a workaround at all. 

It is similar to say, the problem of how to generate a random sorted sequence. There are broadly two approaches: the first tries to generate a random sequence, and filters out the ones that are not sorted. The second approach generates a random sequence, and then sorts it. Clearly the second will be much faster and reliable than the first especially for longer lists.

Retrying has the downside that with more retries, the test just becomes slower and the user is none the wiser.

The old `Command` showed the length of the generated sequences as part of the default output. I think that is a better way to deal with showing potential problems of this kind to the user: https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Commands.fs#L120-L122

In general for a long time I have been on a crusade that users should use combinators like `Prop.collect` and `Prop.classify` to understand what they are generating - otherwise you are testing blind.

MarcoMaar:
I like the idea of propagating via the stamps. I did not think about this at all.

I added a UsedSize member to MachineRun, so we can classify aborted test runs(abortion occurs either via StopCommand or if tryWhere can't find an element). Let me know what you think about it.

The PR is a bit messed up because I accidentally merged the branch in-between. I can create a new PR with one clean commit if you would prefer that.

I also realized there is a problem with `Prop.Classify`. I will create an issue for that.

src/FsCheck/StateMachine.fs
```diff
@@ -118,10 +118,12 @@ type Machine<'Actual,'Model>(maxNumberOfCommands:int) =
...
-type MachineRun<'Actual, 'Model> =
-    { Setup : 'Model * Setup<'Actual,'Model> 
-      Operations : list<Operation<'Actual,'Model> * 'Model>
-      TearDown : TearDown<'Actual> } with
+type MachineRun<'Actual, 'Model>(setup, operations, tearDown) =
+    member val internal Setup : 'Model * Setup<'Actual,'Model> = setup with get, set
+    member val internal Operations : list<Operation<'Actual,'Model> * 'Model> = operations with get, set
+    member val internal TearDown : TearDown<'Actual> = tearDown with get, set
+    member val internal UsedSize = -1 with get, set
+    member this.Clone() = this.MemberwiseClone() :?> MachineRun<'Actual, 'Model>
```
kurtschelfthout: Ho there! You made that beautiful immutable record type into a ugly mutable type? There better be some good explanation for this :)

kurtschelfthout:
Don't worry about the PR, I can squash and merge it in GitHub now.

kurtschelfthout:
Perhaps it would be useful to separate the somewhat controversial part (the `UsedSize`) from the non-controversial part (using `classify` to print out the length of the generated operation lists). I am happy to merge the latter.

For the former I would have to do some thinking. The functionality is not controversial but I'm having a counterreaction to the particular implementation here :)

Or happy to continue discussing here, if you don't care about the delay.

MarcoMaar:
I don't mind a delay at all. I think the functionality of propagating aborted sequences makes sense. At least for me the information is useful.

I changed it to a mutable type because I wanted to make sure that no user can change or set the `UsedSize`. I think `UsedSize `should not be exposed. But yeah, I am also not too happy with my implementation.

Do you have an idea how to implement the functionality clean? I will try to come up with something better the next few days.

kurtschelfthout:
Closing due to inactivity.

PR #274, opened 2016-05-29
Renamed last instance of DontShrink to DoNotShrink

ploeh:
As a follow-up to #273, I'm submitting this very small pull request to complete the rename of `DontShrink` to `DoNotShrink`.

I've begun the work of renaming `DontSize` to `DoNotSize` as well, but it turns out to be slightly more complicated to do, so I thought it better to do that in a separate pull request.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -998,7 +998,12 @@ module Arb =
             |> convert bigint int
 
         ///Overrides the shrinker of any type to be empty, i.e. not to shrink at all.
+        [<Obsolete("Renamed to DoNotShrink.")>]
         static member DontShrink() =
+            generate |> Gen.map DoNotShrink |> fromGen
+
+        ///Overrides the shrinker of any type to be empty, i.e. not to shrink at all.
+        static member DoNotShrink() =
             generate |> Gen.map DoNotShrink |> fromGen
             
         ///Try to derive an arbitrary instance for the given type reflectively. 


ploeh:
@kurtschelfthout, are you doing something strange when you accept my pull requests. GitHub reports them merged OK, but Git on my local machine doesn't think the branches are merged:

image

It's possible that _I'm_ the source of the error, but I don't think I'm doing anything differently than when I'm submitting pull requests for other projects, and there, I typically don't have that problem.

It's not a big problem, though...

kurtschelfthout:
Yes, I am effectively rebasing using GitHub's new "squash and merge" functionality. From my perspective, this keeps the history much more readable. In Git, you should see your commit show up as if you committed and pushed it to master directly.

If a PR consists of just a single commit I've started doing this by default fairly recently. This functionality was only recently introduced by GitHub, and while it was possible to rebase in Git manually I never felt the benefit was worth that overhead; but now it's just a single button in GitHub so the cost is essentially zero.

If the PR contains multiple commits by a single author I'll make a call on a case by case basis, leaning towards merging them as usual, unless I feel the PR contains a bunch of experiments etc  that can be reasonably "squashed away".

If the PR contains multiple commits by multiple authors, I will never squash, because only the merger and one of the authors gets credit on the single commit, and that is obviously unfair.

https://help.github.com/articles/about-pull-request-merge-squashing/

ploeh:
For a couple of years, after @ecampidoglio kindly migrated the AutoFixture code base to GitHub, I deliberately rebased pull requests on _master_, because I was concerned about pull request branches 'polluting' the commit history. That was before I really understood how Git works.

As AutoFixture began to get more and more pull requests, I realised how wrong that was. `git rebase` is, as far as I can tell, never the appropriate way to accept a pull request.

Perhaps you already know this, but some people seem to not realise it: if you do a `git merge` from the command-line, it'll only add a merge commit if it has to. If it can apply the merge by fast-forwarding, it'll do that instead. Some Git management tools, such as _Stash_, however, still insist on adding a merge commit, even when it's unnecessary. I don't know if GitHub does that as well.

For that reason, I always pull down changes and merge from the command-line on my own machine. That also gives me an extra verification step that things still work on my machine. For the projects I administer, I consider that an important quality of the code base.

When it comes to squashing commits, on the other hand, that's a slightly different manner. The problem here isn't that a merge happens, but that history is being rewritten. When it comes to FsCheck, it's your repository, so it's up to you, but personally, I never understood the desire to throw away information. In most enterprises, I fight long and hard battles to get developers to make _smaller_ commits, so I find it a bit ironic that `git squash` even exists...

Anyway, the problem, the slight nuisance that it is, is that since history is being rewritten, my local Git repository doesn't consider the pull request branch merged, and that means extra manual bookkeeping on my part, in order to clean up old branches that actually were merged, even if it doesn't look it.

It's not a big thing, but I wish GitHub had never introduced that _squash and merge_ feature.

ploeh:
...but anyway: thank you for explaining what you do. Now I at least understand why things look like they do from my end. Before, I was worried that something was wrong with my repo, or the way I used it.

kurtschelfthout:
Commits on master are enduring and have a different quality to them. You want your commits to be small, but _not smaller_. In particular, you want the commits on master to tell the story of what happened to your code base straightforwardly. Also, you want each commit to be revertible.

Just like history is not a bland enumeration of facts ("objectively what happened" ) but trying to make sense of the facts, adding connections and a story, an overly focus on keeping things in the repo "objectively as they happened" doesn't necessarily make the most sense.

You are right that if git doesn't have to add a merge commit, it won't. But in a relatively fast moving code base  (and the last few weeks in FsCheck have seen the highest rate of commits so far - a good thing!) a merge commit is in practice frequently necessary, because multiple PRs will have started from the same commit on master. A rebase as such of a single commit doesn't rewrite history; it copies a commit from a branch A to master. The end effect is that your history doesn't end up looking like Guitar hero:

https://twitter.com/HenryHoffman/status/694184106440200192

It's useful for me to write up the release notes that I don't see "did X" - merge - "did Y" - merge and so on. For me, this tells the story more clearly.

I could also ask the PR author to rebase their branch to latest, but then the end effect is the same and now I've made people do more work for the same result.

As for squashing, I use and encourage the use of branches for experimentation and sharing - why should the code on a branch even build? Maybe I just want to show a rough idea to another developer. But if the branch ends up being merged to main, I'd rather rebase it to a set of commits (not necessarily a single one) that each make sense _on master_.

For FsCheck, it sometimes happens that there is some back and forth on a particular PR and I might feel that it's not very useful to have the full history of what happened there in the git history. Do I really need to know that someone corrected a typo?

As I said, this was not particularly high on my list as something to have, but now that it's convenient, I am happy to use it.

ploeh:
@kurtschelfthout, thank you for taking the time to explain your rationale. It makes a lot of sense.

Although I'm afraid I've annoyed you with my original question, I'm still happy I brought this up, because it gave me the opportunity to see things from a different perspective :smile: 

Your analogy to 'real' history was useful to me; it helped me understand that your motivation and priorities are slightly different than mine. Sometimes, I have a tendency to paint myself into a mental corner, in the sense that I convince myself that there's only one correct way to do things, and while I _know_ that that's wrong, it's sometimes difficult to avoid. Learning about the trade-offs made by other people help me realise that my way of working is _also_ a result of trade-offs :flushed: 

I think I now better understand why you manage FsCheck the way you do. You've had great results so far, so keep up the good work :+1: I'll happily contribute again, when I find something to do :smile:

kurtschelfthout:
Honestly there was no annoyance whatsoever. I sometimes ramble on a bit, that's all :) Your contributions are greatly appreciated!

By the way, don't we all paint ourselves in that corner...I constantly struggle with this...

PR #277, opened 2016-06-02
Emit debug info for compiled lambda Expression

jack-pappas:
I recently upgraded a test project from an older version of FsCheck to the current version and found that dotCover seems to have a problem collecting test coverage information after the upgrade. I searched JetBrains' support site and found there's a known (and expected) issue when generating code at run-time without emitting debug information (i.e., a PDB).

I've made a small change (for non-portable targets only) to use the ``Expression.Compile(DebugInfoGenerator)`` overload to emit debugging information. I haven't tested it yet, but I don't think the change will have any negative impact -- I'm able to build and run FsCheck (including this change) without any issues. I'll try to test the change within the next few days to confirm it has the expected effect of allowing dotCover to collect coverage data for FsCheck-based tests.

### DIFF
--- a/src/FsCheck/Reflect.fs
+++ b/src/FsCheck/Reflect.fs
@@ -66,7 +66,15 @@ module internal Reflect =
                                                       :> Expression)
         let body  = Expression.New (ctor, pars)
         let l     = Expression.Lambda<Func<obj[], obj>> (body, par)
-        let f     = l.Compile ()
+
+        let f     =
+#if PCL
+            l.Compile ()
+#else
+            let debugInfoGen = System.Runtime.CompilerServices.DebugInfoGenerator.CreatePdbGenerator ()
+            l.Compile (debugInfoGen)
+#endif
+
         f.Invoke
 
     /// Returns the case name, type, and functions that will construct a constructor and a reader of a union type respectively


kurtschelfthout:
Thanks - yes, let me know if that works. If it does, I see no problem with merging it.

jack-pappas:
I tried this change today; unfortunately it doesn't seem to have fixed the problem though. Perhaps it's a different issue manifesting in the same way.

The original [ticket](https://dotnettools-support.jetbrains.com/hc/en-us/community/posts/205988489-Unit-Testing-works-but-dotCover-failes-to-run-test-Unrecoverable-Error) I found on JetBrains' support forum linked to vanderkleij/Smocks#16, which gave me the idea to try emitting debug info via ``Expression.Compile``.

Any other guesses as to what might be causing the issue? I'll also open a ticket with Jetbrains support to see if they have any ideas (e.g., FsCheck is triggering some bug within dotCover).

kurtschelfthout:
Unfortunately not…if it works with an older version of FsCheck perhaps a git bisect will narrow it down?

jack-pappas:
I have an older version of dotCover on my home machine and was able to run some tests using a recent version of FsCheck without any issues. I think this is probably a dotCover bug so I'll file a bug report with Jetbrains and go from there.

PR #281, opened 2016-06-06
Eliminate tuple allocations in splittable-RNG functions.

jack-pappas:
These changes move the Rnd struct out into the FsCheck namespace and
moved the mixing functions from the Random module into members of the
Rnd struct. The Random module was converted into a static class and the
functions converted to static methods. These methods have been modified
to use 'out' parameters to return the updated Rnd value instead of
returning it with the generated random value in a tuple, thus
eliminating a large source of temporary allocations when running
FsCheck tests (which caused slowdowns due to GC).

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -10,14 +10,42 @@
 
 namespace FsCheck
 
-type internal IGen = 
+open System.Runtime.InteropServices
+
+/// <summary>
+/// 
+/// </summary>
+/// <typeparam name="T">The type of the generated value.</typeparam>
+/// <typeparam name="TState">The type of the generator's state value.</typeparam>
+[<Struct>]
+type GeneratedValue<'T, 'TState> (value : 'T, newState : 'TState) =
+    //
+    member __.Value = value
+    //
+    member __.UpdatedState = newState
+
+//
+[<Interface>]
+type internal IGen =
+    //
     abstract AsGenObject : Gen<obj>
-    
+
 ///Generator of a random value, based on a size parameter and a randomly generated int.
-and [<NoEquality;NoComparison>] Gen<'a> = 
-    private Gen of (int -> Rnd -> 'a * Rnd)
+and [<NoEquality;NoComparison>] Gen<'a> =
+    private Gen of (int -> Rnd -> GeneratedValue<'a, Rnd>)
         ///map the given function to the value in the generator, yielding a new generator of the result type.
-        member internal x.Map<'a,'b> (f: 'a -> 'b) : Gen<'b> = match x with (Gen g) -> Gen (fun n r -> let s,r =  g n r in f s,r)
+        member internal x.Map<'a,'b> (f: 'a -> 'b) : Gen<'b> =
+            let (Gen g) = x
+            Gen (fun size rngState ->
+                // Use the generator to generate a random value.
+                let generatedValue = g size rngState
+
+                // Apply the mapping to the generated value.
+                let mappedValue = f generatedValue.Value
+
+                // Return the generated+mapped value and the updated PRNG state.
+                GeneratedValue (mappedValue, generatedValue.UpdatedState))
+
     interface IGen with
         member x.AsGenObject = x.Map box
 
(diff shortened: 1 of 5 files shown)


jack-pappas:
If these changes seem reasonable and are accepted, I have some follow-up changes in the 'replace-random-noalloc' branch in my fork to eliminate some additional allocations caused by tuple allocations in the Gen functions.

kurtschelfthout:
Thanks, give me some time to take a look.

ploeh:
These are breaking changes, aren't they?

kurtschelfthout:
@ploeh This is for v3; so yes breaking but it doesn't matter.

kurtschelfthout:
Nice. Or rather, it's actually quite ugly! :) But all is forgiven if it's faster. Forgive my laziness,, does it make much of a difference in practice? E.g. if you run the tests, are they faster before/after this change?

I was also playing around with a more stream-like approach, where the seed would be a mutable field on the generator, also to avoid the allocations. Haven't quite figured it out yet.

jack-pappas:
I haven't yet tested the change to see how much of a difference it makes in practice. I'll try to get some rough before/after numbers within the next few days.

The impetus behind this proposed change: I wrote some FsCheck tests recently and noticed one of them was running very slowly. I re-ran the test under a memory profiler and found it was generating a *huge* amount of garbage while running, so there was a significant overhead/bottleneck due to all the Gc'ing involved. One of the larger culprits was tuple allocation in the RNG, so I figured eliminating those allocations might improve the runtime of the test.

kurtschelfthout:
Did you use a custom build to check that? Benause this type (I mean where the Gen returns a tuple) is not in any released build yet.

jack-pappas:
No, I ran the memory profiling against the latest NuGet package. There are actually two places where the tuple allocation is occuring (related to RNG) within the v3 `replace-random` branch:

- Within the RNG itself, because of the way it returns the generated random value and the updated RNG state. These allocations are eliminated by this PR.
- Within some of the `Gen` module functions, related to the way they split RNG states. This source of tuple allocations are present in both `HEAD` and this `replace-random` branch. I've eliminated these too, within the `replace-random-noalloc` branch in my fork, and was planning to submit that as a separate PR (although I'd be happy to merge those changes into this PR if you'd prefer). These changes are breaking changes to the Random and Gen APIs, so I think it makes the most sense to include them in FsCheck 3.x rather than trying to merge the code into `master` (FsCheck 2.x).

jack-pappas:
I've update the PR by merging in my follow-up changes to finish eliminating the tuple allocations. (I was going to submit these as a separate PR but I think it's fine to submit them together now since they're still separate commits.)

I'm not sure why the travis-ci build failed (due to timing out) once I merged in the second commit, especially given the AppVeyor build passed. I'll try running the FsCheck build on Linux+Mono within the next few days to see if there's an actual issue there, or if it was just something transient.

I built the code from this PR and used it to run my tests under the profiler again. There was a significant reduction in the number of allocations (see below for screenshots) and the tests appear to run faster (both with and without the profiler), although I don't have a good way to quantify the speed improvement.

---
NOTE: In the screenshots below, I let my FsCheck tests run for different lengths of time, so the absolute number of bytes/objects allocated aren't directly comparable. Instead, look at the relative number of bytes/objects allocated for the various types (note that the types appear in the same order in all of the screenshots).

FsCheck 2.x allocations
fscheck 2 x allocations cropped

FsCheck 3.x allocations (before applying this PR)
fscheck 3 x replace-random cropped

FsCheck 3.x allocations (after applying this PR)
fscheck 3 x replace-random-noalloc cropped

kurtschelfthout:
Interesting, thanks. I have restarted the travis build just to check if it's reproducible. (Not that the travis build is unreliable. Just checking.)

kurtschelfthout:
Interesting, passed now...that may be worse actually because now I don't know if we were just lucky ;)

jack-pappas:
The Travis-ci build that failed earlier was due to a timeout. I can't find the failed results now, but IIRC the build time for that one was several times as long as the AppVeyor build. The build times for the most recent builds are closer and can likely be chalked up to differences in the specs of the VMs being used (or CLR vs Mono). I think the previous failure may have been a fluke, and if it wasn't, I don't think there's any way we'll be able to tell.

kurtschelfthout:
Yeah, I saw it was a timeout. I've seen infinite loops before so I was a bit cautious. Agree - I don't think there is much stopping this from being merged.

Issue #283, opened 2016-06-07
Prop.Classify does not classify all testruns


MarcoMaar:
The title says most of it. 
The following code
```csharp
Prop.ForAll<int>(i => true).Classify(true, "all").QuickCheck();
```
returns
`Ok, passed 100 tests (99% all).`

I did not have time to debug properly yet but this is what I found out.
https://github.com/fscheck/FsCheck/blob/6369b0d57daba4112fc9dd954c1085f6c31408be/src/FsCheck/Runner.fs#L177
When the sequence is folded after execution the last element is still set to `Generated `and not to `Passed`. Therefore the last stamp is not collected.

Update: see PR #284

MarcoMaar:
closed with #284

PR #284, opened 2016-06-07
Prop.Classify

MarcoMaar:
PR related to #283 

`Prop.Classify` did not classify the last element:
Reason:
`takeWhile `does not include the element which does not satisfy the
predicate.
Therefore, changed `takeWhile `to a custom implementation

I added the implementation to the Common module. Not sure if it belongs here.
The solution was taken from a stackoverflow answer
http://stackoverflow.com/a/12564899/2219351

### DIFF
--- a/src/FsCheck/Common.fs
+++ b/src/FsCheck/Common.fs
@@ -49,3 +49,24 @@ module internal Common =
 
     let (|MapContains|_|) key value =
         Map.tryFind key value
+
+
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
It seems fine, just wondering what happened with the AppVeyor build on this PR...doesn't seem to have been kicked off. I'll take a look.

PR #285, opened 2016-06-08
Move PCL-specific culture names to Data module

ErikSchierboom:
In #255, the list of top-level domains was placed in a separate `Data` module to prevent them from cluttering the `Arb` module. This PR does the same thing for the PCL-specific culture names list. It makes the `Arb` module read better IMHO.


ploeh:
LGTM :+1:

Issue #286, opened 2016-06-09
Experimental.StateMachine Disable Shrinking not possible


MarcoMaar:
Disabling(or Changing) Shrinking is currently not possible since we are using the Default Shrinker.
https://github.com/fscheck/FsCheck/blob/c3578323ad153be891fb3b4eb2b4fabbc20a96d4/src/FsCheck/StateMachine.fs#L294

I will create a PR and fix it.

PR #287, opened 2016-06-09
Enable registering of custom shrinker for StateMachine

MarcoMaar:
PR for #286


MarcoMaar:
I have a problem with the testcase that I have written for this issue.

As you can see I register a custom shrinker and then "restore" the default shrinker. This has to be done in order for the other test to work.

However, I think this solution is pretty ugly... Any idea how we could write a cleaner test?

kurtschelfthout:
I dislike this approach, I'm trying to get rid of the side-effecting registration of generators/shrinkers in v3. I think we should put a shrink method on `Machine` instead.

kurtschelfthout:
Closing per above comment.

Issue #282, opened 2016-06-07
Allow to run a suite of tests with the same seed


theimowski:
I have a somewhat unusual case for using FsCheck, as I once described [in my blog post](http://theimowski.com/blog/2015/12-21-property-based-testing-xslt/index.html).

Now, while I found it convenient to use `[<Property>]` attribute for my properties to nicely integrate with XUnit, I discovered that test run performance was very poor due to fact that most of time was spent in the XSLT processor (N tests multiply 100 runs = 100*N XSLT processor invocations).
I've workaround this issue by using a custom sub-type of `Property` attribute:

```fsharp
type XsltProperty() =
    inherit FsCheck.Xunit.PropertyAttribute(Replay  = XsltProperty.Seed)
    
    static let seed = 
        let (FsCheck.Random.StdGen(a,b)) = FsCheck.Random.newSeed()
        sprintf "%d,%d" a b

    static member Seed with get() = seed
```

Because `Seed` member is static, its value is set upon first invocation, and all functions marked with `[<XsltProperty>]` will always run with the same seed (the seed will differ between one run and another).
Thanks to this, I was able to **memoize** XSLT processing for a specific XML document, and thus reduce number of XSLT processor invocations from 100*N to 100.

I'm wondering whether it might be worth adding something similar to the FsCheck itself (maybe only to the XUnit plugin?), which would allow to run a set of tests with the same seed. 
What you think?

kurtschelfthout:
Not to diminish the value of your particular case, but it sounds like it adds too limited general purpose value. To do this I think we'd have to add a new attribute that users can put on the module or type level - that sounds like a bit too much for this one feature. Perhaps though we could add a more general configuration attribute where we could set all the configuration as on the `PropertyAttribtue`, but it would take effect on all the tests in the type/module. But then we'd also have to make sure that the `PropertyAttribute` on the method level properly overrides these.

moodmosaic:
>Perhaps though we could add a more general configuration attribute where we could set all the configuration as on the PropertyAttribtue, [..]. But then we'd also have to make sure that the PropertyAttribute on the method level properly overrides these.

Doesn't it sound a bit complicated?

---

IIRC, there's already a *Replay* functionality which is set to `None` by default, and setting it to `Some StdGen` will make the test to be replayed in the exact same way.

theimowski:
That's right - I'm not sure whether my use case is worth the hussle implementing such configuration in FsCheck itself.
[...]

PR #288, opened 2016-06-15
XUnit properties - allow configuration on module level

theimowski:
**Marking with WIP prefix** to get early feedback

This is what I came up with regarding #282.
Unfortunately the change turned out to be quite big.

The basic idea is to allow configuring properties on class/module level, and override the config on property level.
For that purpose I've created internal `PropertyConfig` type which represents values of `PropertyAttribute`.
In addition to that, I've made a derived class from `PropertyAttribute` called `GeneralAttribute` (name should probably be changed).
The subclass relation is there to remove code duplication and maintain .net Attribute requirements (public automatic properties)

One of drawbacks is that with these changes, there would be 2 possible ways to add `Arbitrary` types on class/module level: with `ArbitraryAttribute` and `GeneralAttribute`

Any comments welcome - if you decide that the change is too big / breaking and should not be merged, I'll be fine with that as well.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -89,7 +89,8 @@ Target "AssemblyInfo" (fun _ ->
           Attribute.Description package.Summary
           Attribute.Version release.AssemblyVersion
           Attribute.FileVersion release.AssemblyVersion
-        ] @ (if package.Name = "FsCheck" then [Attribute.InternalsVisibleTo("FsCheck.Test")] else []))
+        ] @ (if package.Name = "FsCheck" || package.Name = "FsCheck.Xunit"
+             then [Attribute.InternalsVisibleTo("FsCheck.Test")] else []))
     )
 )
 
--- a/docs/content/RunningTests.fsx
+++ b/docs/content/RunningTests.fsx
@@ -137,9 +137,8 @@ let ``abs(v) % k equals abs(v % k)`` v (NonZeroInt k) =
     (abs v) % k = abs(v % k)
 
 (**
-Likely on of the most useful configuration options of `PropertyAttribute` is the ability to register or override an `Arbitrary`
-instance just for that test method. You can also use the `ArbitraryAttribute` to register or override `Abritrary` instances on 
-a per class or per module basis. For example:*)
+Likely one of the most useful configuration options of `PropertyAttribute` is the ability to register or override an `Arbitrary`
+instance just for that test method. You can also use the `PropertiesAttribute` (note the plural form) to set custom configuration options per class or per module basis. For example:*)
 
 type Positive =
     static member Double() =
@@ -151,8 +150,8 @@ type Negative =
         Arb.Default.Float()
         |> Arb.mapFilter (abs >> ((-) 0.0)) (fun t -> t <= 0.0)
 
-[<Arbitrary(typeof<Negative>)>]
-module ModuleWithArbitrary =
+[<Properties( Arbitrary=[| typeof<Negative> |] )>]
+module ModuleWithProperties =
 
     [<Property>]
     let ``should use Arb instances from enclosing module``(underTest:float) =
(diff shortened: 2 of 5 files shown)


kurtschelfthout:
Sorry for the delay, RL is catching up to me. It looks good in broad strokes; I haven't had a chance to review in detail yet. If you would like a more detailed review before continuing, let me know.

theimowski:
Yes, I'd be grateful for detailed review, specially around:

* `GeneralAttribute`, `PropertyAttribute` - names of attributes, are we fine with the inheritance relation
* the module-level `ArbitraryAttribute` - should it be marked as obsolete? (because `GeneralAttribute` lets to set the `Arbitrary` member as well)

ploeh:
I must admit that I find it difficult to follow this suggested change :flushed: 

Could you show a what usage would look like?

theimowski:
Sure.
The original intent was, as I stated above, issue #282 
There came a suggestion from @kurtschelfthout to generalize the concept, and instead of specifying only `Replay` on module level, we would be able to set all other configuration options on module level.

Given the following example:

```fsharp
[<General(MaxTest = 10, EndSize = 10)>] 
module Properties

[<Property>]
let ``property 1`` input =
    true

[<Property(MaxTest = 500)>]
let ``property 2`` input =
    true

[<Property(Replay = "123,456")>]
let ``property 3`` input =
    true
```

* `property 1` would use default config + overriden `MaxTest = 10` and `EndSize = 10` from `General` attribute
* `property 2` would use default config + overriden `EndSize = 10` from `General` attribute  and `MaxTest = 500` from `Property` attribute
* `property 3` would use default config + overriden `MaxTest = 10` and `EndSize = 10` from `General` attribute and `Replay = "123,456"` from `Property` attribute

Does it make any sense?

ploeh:
Instead of `[<General>]`, how about enabling users to apply the `[<Property>]` attribute on the module level as well? Or perhaps call the attribute `[<Properties>]`?

Without a namespace, the type name `General` seems awfully vague...

theimowski:
I agree, TBH I don't like the `General` name as well, but couldn't come up with anything reasonable. `Properties` sounds good for me

kurtschelfthout:
First of all, apologies for the extremely long time it has taken me to get to this.

The code looks good; I have some general remarks.

- I agree `General` is not a good name. `Properties` is better, but doesn't quite hit the spot either :) `PropertyConfig`? `FsCheckConfig`?
- Some tests of the configuration would be nice; there are some tests around the existing `Arbitrary` attribute.
- We should obsolete the existing Arbitrary attribute and point users to this instead.

theimowski:
Thanks, I've applied suggested changes. `Properties` is fine for me.
Shall I come up with more tests? Not sure how can I meta test configuration for a property - any tips?
Are there any docs which should be updated as well?

kurtschelfthout:
Nice. Perhaps we can make FsCheck.Xunit `InternalsVisibleTo` FsCheck.Test, so you can test the internal `PropertyConfig` module more easily?

You can also always set a seed and a MaxTest = 1 and then check that a generated value (say an `int`) is the same one every time (and so also that you also run one test). And perhaps you could vary where you put that configuration like with the arbitrary stuff (e.g. put another seed on the module level than on the function level, function should take precedence etc).

moodmosaic:
>Perhaps we can make FsCheck.Xunit `InternalsVisibleTo` FsCheck.Test, so you can test the internal `PropertyConfig` module more easily

Why not just make it [public (but hidden in plain sight)](http://blog.ploeh.dk/2015/09/21/public-types-hidden-in-plain-sight/) then?

kurtschelfthout:
@moodmosaic Because it's not public functionality. A user can't do anything with it. But using `InternalsVisibleTo` we can write some nice unit tests for something that is otherwise annoying to test.

theimowski:
@kurtschelfthout I tried to test a single case, however the results differ. Can you tell what's wrong with the below code?

```fsharp
module Test =
        [<Property( MaxTest = 1, StartSize = 100, EndSize = 100, Replay = "12345,67890")>]
        let ``should use configuration on method preferentially``(x:int) =
            x =! Gen.eval 100 (Random.StdGen(12345,67890)) Arb.generate<int>

```

This gives following:

```
 FsCheck.Xunit.PropertyFailedException :
      Falsifiable, after 1 test (1 shrink) (StdGen (12345,67890)):
      Original:
      18
      Shrunk:
      0

      ---- Swensen.Unquote.AssertionFailedException : Test failed:

      0 = -82
      false
```

kurtschelfthout:
Right; just use a hardcoded value `18` in this case. The two ways you're comparing now use slightly different code paths (reflective call via function vs more direct Gen.eval), so there is an extra step somewhere leading to the difference in the generated value.

theimowski:
like [this](https://github.com/fscheck/FsCheck/pull/288/files#diff-285222947e56e5faa81a9962c884a5bdR124)?

theimowski:
ok I've added a couple of tests for internal `PropertyConfig` functions - are those satisfactory or is there anything else you'd like to see?

kurtschelfthout:
This is great. When the tests come back I'll be happy to merge!

theimowski:
cool thanks!

PR #289, opened 2016-06-22
when you have a module of property based tests, and no zero param exa…

sgtz:
fix for when you've only got property-based tests in a module, and when Arb.registerByType gets called (when manually invoked via FSI).  It looks to me that the = 0 params guard is no longer needed.  Please confirm if I am reading this correctly.


kurtschelfthout:
It is apparently needed, as this causes a failing test.

Can you reproduce the problem you are seeing? From your description I don't quite understand what is going on.

sgtz:
clearly this can't go in yet due to failed test.  Please advise if you think this is still a good way to go.

PR #291, opened 2016-07-17
Optimizations for array generation and shrinking

jack-pappas:
Modified Gen functions for generating arbitrary array instances to avoid intermediate conversions to F# lists (and resulting allocations).
Modified shrinker for 1D arrays to avoid intermediate list allocations.
Modified shrinker for strings to use the array shrinker, and subsequently avoid allocating intermediate char lists.

I noticed these allocations when profiling a slow-running FsCheck test I wrote at one point (for a data type which used both arrays and strings). Running ``build.cmd`` before and after this change seems to show the tests running slightly faster, although I haven't repeated it enough times for statistical significance. In any case, tests which heavily use arrays and strings should notice a performance improvement due to eliminating the conversions to/from F# lists (and also due to spending less time in GCs related to those list allocations).

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -10,6 +10,10 @@
 
 namespace FsCheck
 
+// Disable warnings about calling certain FsCheck functions from F#.
+// Using them internally within FsCheck is important for performance reasons.
+#nowarn "10001"
+
 open System
 
 #if PCL
@@ -468,7 +472,7 @@ module Arb =
             let shrinker (s:string) = 
                     match s with
                     | null -> Seq.empty
-                    | _ -> s.ToCharArray() |> Array.toList |> shrink |> Seq.map (fun chars -> new String(List.toArray chars))
+                    | _ -> s.ToCharArray() |> shrink |> Seq.map (fun chars -> new String(chars))
             fromGenShrink (generator,shrinker)
 
         ///Generate an option value that is 'None' 1/8 of the time.
@@ -506,11 +510,13 @@ module Arb =
             { new Arbitrary<list<'a>>() with
                 override __.Generator = Gen.listOf generate
                 override __.Shrinker l =
-                    match l with
-                    | [] ->         Seq.empty
-                    | (x::xs) ->    seq { yield xs
-                                          for xs' in shrink xs -> x::xs'
-                                          for x' in shrink x -> x'::xs }
+                    let rec shrinkList l =
+                        match l with
+                        | [] ->         Seq.empty
+                        | (x::xs) ->    seq { yield xs
+                                              for xs' in shrinkList xs -> x::xs'
+                                              for x' in shrink x -> x'::xs }
+                    shrinkList l
             }
 
         ///Generate an object - a boxed char, string or boolean value.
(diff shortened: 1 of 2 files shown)


src/FsCheck/Gen.fs
```diff
@@ -330,14 +343,38 @@ module Gen =
     ///Sequence the given list of generators into a generator of a list.
     //[category: Creating generators from generators]
     [<CompiledName("Sequence"); CompilerMessage("This method is not intended for use from F#.", 10001, IsHidden=true, IsError=false)>]
-    let sequenceToSeq generators = 
-        generators |> sequence |> map List.toSeq
+    let sequenceToSeq<'T, 'TSource when 'TSource :> IEnumerable<Gen<'T>>> (generators: 'TSource) =
```
kurtschelfthout: This function used to only take a single type argument. So this unfortunately breaks compatibility.

src/FsCheck/Arbitrary.fs
```diff
@@ -555,6 +563,32 @@ module Arb =
...
+        static member String() = 
+            let generator = Gen.frequency [(9, Gen.map (fun (chars:char[]) -> new String(chars)) generate);(1, Gen.constant null)]
+            let shrinker (s:string) = 
+                    match s with
+                    | null -> Seq.empty
+                    | _ -> s.ToCharArray() |> shrink |> Seq.map (fun chars -> new String(chars))
+            fromGenShrink (generator,shrinker)
+
+        ///Generate an object - a boxed char, string or boolean value.
+        static member Object() =
```
kurtschelfthout: This Object() method is just moved, right?

src/FsCheck/Arbitrary.fs
```diff
@@ -555,6 +563,32 @@ module Arb =
                             
             fromGenShrink (Gen.array2DOf generate,shrinkArray2D)
 
+        ///Generates arbitrary strings, which are lists of chars generated by Char.
+        static member String() = 
+            let generator = Gen.frequency [(9, Gen.map (fun (chars:char[]) -> new String(chars)) generate);(1, Gen.constant null)]
+            let shrinker (s:string) = 
+                    match s with
+                    | null -> Seq.empty
+                    | _ -> s.ToCharArray() |> shrink |> Seq.map (fun chars -> new String(chars))
```
kurtschelfthout: Any other change besides removing the toList? (as a side note, diffs are much easier to understand for me if you don't move things around...)

kurtschelfthout:
Overall, I'm liking this PR!

src/FsCheck/Gen.fs
jack-pappas: Ah, didn't think about that - I'll revert this part of the changes. Now that I think about it again, I'm not sure adding the extra type parameter was going to accomplish what I had in mind anyway.

src/FsCheck/Arbitrary.fs
jack-pappas: Yes. I'll revert that part of the change. I must've been on autopilot and thinking of reordering things as if they were let-bindings. They're all static members though so there's no need to reorder them.

src/FsCheck/Arbitrary.fs
jack-pappas: No, no change besides that. Removing the toList just makes it use the shrinker for array instead.

jack-pappas:
I've moved the Arbitrary members back to their original locations and fixed that one incompatibility in Gen. The diff should be smaller and more readable now.

PR #303, opened 2016-08-27
Handled char-based flags enums.

ploeh:
As far as I can tell, char-based flags enums don't make much sense, since
they don't seem to support binary 'or' or 'and'. (If there's a way to 'or'
two char-based flags together, I haven't found out how to do it.) As the
added test demonstrates, however, it's possible to define and compile a
char-based flags enum, so FsCheck ought to explicitly handle such a type.

### DIFF
--- a/src/FsCheck/ReflectArbitrary.fs
+++ b/src/FsCheck/ReflectArbitrary.fs
@@ -45,6 +45,7 @@ module internal ReflectArbitrary =
            elif elementType = typeof<int>    then listOf elems |> map (orElems<int>    t)
            elif elementType = typeof<uint64> then listOf elems |> map (orElems<uint64> t)
            elif elementType = typeof<int64>  then listOf elems |> map (orElems<int64>  t)
+           elif elementType = typeof<char>   then elems // Binary 'or' not defined for char enums
            else invalidArg "t" (sprintf "Unexpected underlying enum type: %O" elementType)
        else 
            elems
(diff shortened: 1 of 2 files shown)


ploeh:
Oh, I see that the Travis build failed. From the log, though, it seems unrelated to the changes in this PR. Could it be an Erratic Tests? Or is there a connection I'm simply not seeing?

kurtschelfthout:
It looks like mono is not happy with (very likely, although xunit apparently has no time to actually output the test name in which this happens) char-based Flags enums...it's throwing an assertion failed in the mono runtime.

ploeh:
It _is_ a weird combination. Is there anything we can do about it (conditional compilation flags or the like), or do you want to ignore the issue (i.e. declining the pull request)?

kurtschelfthout:
There is no infrastructure for conditional compilation mono/.net at the moment. 

As I'm sure you understand, I can't accept the PR as is. So up to you if you want to bite off either getting the mono bug fixed (if it is indeed a bug, well at least I suppose it shouldn't crash like that :) ) or additionally creating the mono/.net compilation flags in FsCheck. I am doubtful the extra complexity of the latter is worth it for this specific case though.

ploeh:
I understand. Since I don't have any means of troubleshooting a mono build, I'm going to close this.

As you hint, I'm not even sure it's a bug. While it makes sense to define a normal enum based on `char` values, and it also makes sense to enable the `[<Flags>]` annotation for various enum types, that particular combination makes little sense.

PR #308, opened 2016-09-08
add osx to travis (on master branch)

enricosada:
with mono `latest`, let's check if osx works

with new edits from maintainers feature of github, i think you @kurtschelfthout  can push commit to this branch


enricosada:
fail on `osx`, ok on `linux`

```
				Tool /Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/4.5/mcs.exe execution started with arguments: /noconfig /debug:pdbonly /optimize+ /out:obj/Release/FsCheck.CSharpExamples.exe Program.cs Properties/AssemblyInfo.cs obj/Release/.NETFramework,Version=v4.5.AssemblyAttribute.cs /target:exe /define:TRACE /nostdlib /reference:/Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/gac/FSharp.Core/4.3.1.0__b03f5f7f11d50a3a/FSharp.Core.dll /reference:/Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/4.5-api/System.dll /reference:/Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/4.5-api/System.Data.dll /reference:/Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/4.5-api/System.Core.dll /reference:/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll /reference:/Library/Frameworks/Mono.framework/Versions/4.4.2/lib/mono/4.5-api/mscorlib.dll /warn:4
Program.cs(143,27): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(143,31): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `Choose'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(146,30): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(146,34): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `OneOf'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(148,31): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(148,35): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `Frequency'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(165,33): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(165,37): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `Choose'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(200,20): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(200,24): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `Sized'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(208,28): error CS0305: Using the generic type `FsCheck.Gen<a>' requires `1' type argument(s)
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
Program.cs(208,32): error CS0117: `FsCheck.Gen<a>' does not contain a definition for `Sized'
				/Users/travis/build/fscheck/FsCheck/src/FsCheck/bin/Release//FsCheck.dll (Location of the symbol related to previous error)
```

baronfel:
I can reproduce this failure on my local Mono install. I'm running mono 4.4.2 stable, installed from Homebrew.

enricosada:
really thx for help @baronfel

enricosada:
The dll built by @baronfel on osx

[FsCheck.zip](https://github.com/fscheck/FsCheck/files/462334/FsCheck.zip)

enricosada:
There is no `Gen` class, only `Gen<a>`, that's explain the error

osx:

image

windows:

image

baronfel:
Well now I've learned about Xamarin Studio's Assembly viewer. I definitely see the same thing.

screen shot 2016-09-08 at 1 05 10 pm

enricosada:
so `GenModule` instead of `Gen`

src/FsCheck/Gen.fs
```diff
@@ -181,7 +181,7 @@ module Gen =
         | :? array<_> as arr ->
             elementsArr arr
         | _ ->
-            choose (0, (Seq.length xs)-1) |> map (flip Seq.nth xs)
+            choose (0, (Seq.length xs)-1) |> map (flip Seq.item xs)
```
kurtschelfthout: As you've noticed, this fails because master is still on an older version of FSharp.Core. :)

kurtschelfthout:
That must be an F# compiler bug I think that somehow manifests only when  you build on OSX. There is no reason why the `Gen` module should have `Module` appended to it (and indeed, this does not happen if you build .NET on Windows, or mono on Linux...).

enricosada:
Probably https://github.com/Microsoft/visualfsharp/pull/1319 and a different version of f# compiler

kurtschelfthout:
Thanks @enricosada -  I've added some comments to the RFC. I don't particularly agree with this new behavior...

enricosada:
RFC or not, i think next questions are:

1. a workaround? `[<CompiledName>]` or something else?
2. why osx has this? it should not, it's using the same mono version of ubuntu.

src/FsCheck/Gen.fs
enricosada: `#ifdef` all the things! or not? i dont like warnings

enricosada: btw i added this change because @baronfel tried to compile ootb with xamarin studio but failed (warning as error?), and the real compiler error of this issue was after these two warnings

ploeh:
> using the same mono version of ubuntu

Not that I'd claim to be an expert at all, but I thought [MacOS was based on Unix](https://en.wikipedia.org/wiki/OS_X) (i.e. _not_ Linux, so also _not_ Ubuntu)...

enricosada:
@ploeh sry, bad phrase.

Using the same mono version number (4.2.2) in osx and ubuntu, why the fsc has that new RFC feature? or is a bug with osx only?
I tried also using mono `4.0.5` on osx (see [travis logs](https://travis-ci.org/fscheck/FsCheck/builds/158522722) ) and the error is the same. why? 
is mono on osx installing HEAD fsc? dunno but is strange and should be investigated

kurtschelfthout:
The correct fix here, imo, is to use the same version (or as close as possible) of the compiler on linux/osx as the one we're using on Windows.

Note that according to the RFC, even `CompiledName` doesn't work to override the name.

enricosada:
@kurtschelfthout @ploeh we are getting this error on travis with mono version 4.0.5 (one year old)

```
Mono JIT compiler version 4.0.5 ((detached/1d8d582 Tue Oct 20 15:15:33 EDT 2015)
```

that's strange..

src/FsCheck/Gen.fs
kurtschelfthout: That means you're compiling against a later version of FSharp.Core. FsCheck on master is compiled against 4.3.1.0 - this was not obsoleted in that version. So there is no reason for the ifdef. FsCheck project indeed has warnings as errors turned on.

kurtschelfthout:
@enricosada I mean version of fsc. Is that controllable?

kurtschelfthout:
Yes, agree that's weird.

kurtschelfthout:
@enricosada can you enable write access on this branch? see https://github.com/blog/2247-improving-collaboration-with-forks

GitHub seems to claim I don't have write access...

enricosada:
Allow edits from maintainers.

Was already checked, dunno you get error in push?

kurtschelfthout:
sorry, my mistake, git trouble :) worked now. I tried a few workarounds, we'll see...

enricosada:
@kurtschelfthout remove some from build Matrix in travis. 
Atm there are lots mono versions...

kurtschelfthout:
This is so weird. The only build on linux that works, which claims to use F# 4.0 compiler and FSharp.Core 4.3.1.0 nonetheless gives a deprecation warning for Seq.nth. But for sure the FSharp.Core 4.3.1.0 on windows does not have that deprecation warning. 

On OSX I have no idea what it's doing but it never works. And old mono on linux seems to kill the mono runtime when running paket.

What a mess!

kurtschelfthout:
I give up on this :)

PR #315, opened 2016-09-29
Add check extensions for xunit

eggapauli:
Use xUnit's `ITestOutputHelper` to output FsCheck messages

### DIFF
--- a/src/FsCheck.Xunit/CheckExtensions.fs
+++ b/src/FsCheck.Xunit/CheckExtensions.fs
@@ -1,0 +1,43 @@
+﻿namespace FsCheck.Xunit
+
+open System
+open System.Runtime.CompilerServices
+open FsCheck
+open Xunit.Abstractions
+
+module private Helper =
+    let private runner (testOutputHelper: ITestOutputHelper) =
+        { new IRunner with
+            member __.OnStartFixture t =
+                Runner.onStartFixtureToString t |> testOutputHelper.WriteLine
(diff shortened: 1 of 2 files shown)


eggapauli:
I wasn't sure. I mean I couldn't find a project that tests the xUnit attribute so I thought it's not necessary to include tests. If an inline example is enough, this can be used like this (beware of typing errors):

    public class Test {
        private readonly ITestOutputHelper _TestOutputHelper;
        public Test(ITestOutputHelper testOutputHelper) {
            _TestOutputHelper = testOutputHelper;
        }
        [Fact]
        public void Test1() {
            Prop
                .ForAll(Arb.Default.Int32(), i => ...)
                .VerboseCheckThrowOnFailure(_TestOutputHelper);
        }
    }

A short explanation: xUnit 2 doesn't capture messages written to the console because tests can be executed in parallel so it can't really tell which message belongs to which test. That's what `ITestOutputHelper` tries to solve. This interface has a single method `WriteLine` and xUnit will automatically pass it in as a constructor argument. This PR makes it possible to pass an `ITestOutputHelper` to FsCheck so that FsCheck messages are captured by xUnit.

kurtschelfthout:
Neat. Don't think tests are really possible or necessary here but please do consider adding a note or something to the documentation, say here: https://github.com/fscheck/FsCheck/blob/master/docs/content/RunningTests.fsx

PR #320, opened 2016-10-03
Corrected an error with conditional compilation

ploeh:
This is very embarrassing, but as far as I can tell, I made an error in my last commit. The intent was to
make HostName available on both the full framework and PCL, but I think I got it wrong. I hope this commit fixes it.

**Given that my previous pull request demonstrates that I'm having trouble with conditional compilation, an extra-critical code review would be in order here!**

Sorry about the mess :flushed: 

The call to ToCharArray() is necessary because, apparently, System.String doesn't implement `char seq` on all portable platforms... At least, without it, compilation fails on PCL.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -984,16 +984,14 @@ module Arb =
                 |> Seq.map IPAddress
         
             fromGenShrink (generator, shrinker)
-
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
No worries, I should've looked more carefully before accepting :)

PR #321, opened 2016-10-03
Fixed regression in VB.NET support

pblasucci:
This addresses a regression to the fix made in #134. The regression seems to be caused by FAKE stomping on the `AssemblyInfo.fs` file (where the code lifed previously). So, as a fix, the code (a 2 lines of it), is moved to a different file (one that doesn't get rewritten as part of the build process). Additionally, the commit includes a code comment, since the code looks a bit out-of-place in it's new location.

### DIFF
--- a/src/FsCheck/Common.fs
+++ b/src/FsCheck/Common.fs
@@ -70,3 +70,11 @@ module internal Common =
         // (making sure that the enumerator gets disposed)
         seq { use en = s.GetEnumerator()
             yield! loop en }
+
+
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Whoops. Thanks for fixing again! I will release this ASAP, which may be tonight but more likely tomorrow.

pblasucci:
Awesome! Thanks! (and no worries... I'm sorry I didn't realize the regression sooner).

PR #323, opened 2016-10-05
Overrode HostName.ToString()

ploeh:
Since HostName is only a wrapper over string anyway, it'd be reasonable to
override ToString() to return the string in question.


moodmosaic:
Looks good to me 👌

Issue #332, opened 2016-12-06
DoNotSize generators are skewed


ploeh:
It looks like some of the DoNotSize generators are skewed:

```
> Arb.generate<DoNotSize<int64>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(-1, 988696); (1, 11304)]

> Arb.generate<DoNotSize<uint64>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(0, 359391); (1, 640609)]

> Arb.generate<DoNotSize<int32>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(1, 371659); (-1, 628341)]

> Arb.generate<DoNotSize<uint32>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(1, 500065); (0, 499935)]

> Arb.generate<DoNotSize<int16>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(1, 499986); (-1, 500002); (0, 12)]

> Arb.generate<DoNotSize<uint16>>
    |> Gen.map (fun (DoNotSize x) -> float x)
    |> Gen.sample 0 1000000
    |> List.countBy sign;;
val it : (int * int) list = [(1, 999985); (0, 15)]
```

The 16-bit generators both look good, the 32-bit generators are suspect, and the 64-bit generators are skewed by orders of magnitude.

I expected these generators to generate values drawn from a uniform distribution across their entire range, but that seems not to be the case. Is this on purpose? If so, why?

kurtschelfthout:
I'm afraid this may reflect the "badness" of the current random number generator - and hence why there is a replacement on the fscheck3 branch.

The replacement generates uniform numbers (at least, the authors of the paper I implemented ran a decent battery of statistical tests on the output) within the full int64 range. The current implementation generates _untested_ uniform output in _almost_ the int32 range. Afaik there is paper by Koen Claessen et al that shows some pretty skewed results with this random generator.

[...]

PR #333, opened 2016-12-07
Adjusted documentation for DoNotSize Arbitraries

ploeh:
Per the discussion in #332, these Arbitraries do not generate values
uniformly distributed across their entire range, so the comments (that
become documentation) should not claim that.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -354,14 +354,14 @@ module Arb =
             Default.Int32()
             |> convert (abs >> uint32) int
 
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
I agree with the spirit, but as written now I feel it's just needlessly verbose:

    Generate arbitrary uint32 that is distributed in the range of uint32 values.

The first three words already say it's distributed from the range of unint32 values...

Maybe we should say something to indicate these generators are _unrestriced by size_? E.g.

    Generate arbitrary uint32 that is unrestricted by size.

ploeh:
OK, PR updated 😄

Issue #98, opened 2015-05-14
Support NUnit 3


jhamm:
Currrently NUnit v3 is in Beta. I have the updates to support it in the my branch [jhamm/FsCheck](https://github.com/jhamm/FsCheck) in the support-nunit3 branch.

The FsCheck.NUnit.Addins project was not needed to support NUnit 3, so I removed it.

The branch is not ready to be merged into the main branch because it is still in beta and not all of the related projects, like [NUnit.Runners](https://www.nuget.org/packages/NUnit.Runners/), support NUnit 3 yet. For testing, I built and changed the references to pointed the references to the local copies of the dll's.

kurtschelfthout:
Cool, would be great to get rid of that stuff. It's been causing a lot of issues.

kurtschelfthout:
Alright the PR is now merged and cleaned up all the stuff I expected it to clean up, so as soon as NUnit3 hits the street this is going out the door too.

Meanwhile, please test like crazy via the AppVeryor feed, 2.1.0-b296 and report any issues here.

kkm000:
Hi @kurtschelfthout, NUnit 3 is said to be in the latest stages of beta, and production-ready for all practical purposes. Would it be very hard to publish a pre-release package of `FSCheck.NUnit` on NuGet? If you do not want to do that for any reason, then how can I build a package for our internal use--is the tip of the nunit3 branch the correct source?

kurtschelfthout:
Yes, that's currently the last one, but I should merge from main and change the version. That version won't have xunit2 support in it. Don't know if you care. I think it should be good enough to test NUnit 3 for now. I'll update this thread when I have an updated version.

kkm000:
Thanks, that will do it for me. I noticed that later beta labels did not work, only b296. So far looks good (VS2015).

kurtschelfthout:
2.2.0-nunit3-b324 on AppVeyor is the latest version - I merged in the changes we did in 2.1 there.

kkm000:
Thanks, switching to it.

[...]

PR #337, opened 2017-01-13
Make tests run under nunit3-vs-adapter and nunit-gui (for Issue #98).…

marklam:
… Also seems to fix running under NCrunch 2.27.03

The parent suite for the method wasn't being preserved in FsCheckTestMethod, and invoking the run command created a new (unexpected) test method instead of running the previously created one.
(Originally raised as https://github.com/nunit/nunit3-vs-adapter/issues/293 but was not a bug in the nunit adapter.)


kurtschelfthout:
Super! I'm going to double-check before releasing but from the looks of it this astronomically increases the chance that we finally upgrade the adapter to nunit 3 :)

PR #339, opened 2017-01-14
Quickstart with F#-native runner, expecto

haf:
Needs CI.


kurtschelfthout:
I'm pretty sure that despite all the green this doesn't work because expecto is not in packages. (FSharp.Documentation just ignores code it can't execute, or something, so it doesn't show up in CI.) Should be easy to fix...

PR #346, opened 2017-02-08
add a regression test agains 344 (stackoverflow)

0x53A:
If you have any idea how to better abort the run after a few iterations ...

This fails against master, but passes against my pr.


kurtschelfthout:
Thanks! Can you just add this commit to the other PR with the fix, and move the test into FsCheck.Test/Runner.fs?

0x53A:
done

Issue #83, opened 2015-01-25
Support for running async tests in parallel?


polytypic:
I would like to test timing properties of concurrent code.  The tests take very little CPU time, but lots of wall-clock time if you run them sequentially.  The tests are such that it should be possible to run tens of thousands of tests&mdash;several orders of magnitude more than the number of cores&mdash;in parallel without problems.  Have there been attempts to allow FsCheck to run async tests in parallel?

kurtschelfthout:
I did a half-hearted attempt a long time ago. Requires a bit of admin to shuffle things around, don't think it's fundamentally hard. I probably made a bit of a mess with the whole runner setup, to be honest. (I mean, the code in Runner.fs specifically, which would need to be modified) It looks more convoluted than its worth, e.g. using the IEnumerable to run the tests. But never got round to cleaning it up.

I think fundamentally it's in decent shape to do this kind of thing; the random seed can be split (though the random properties of these split seeds have never been proven afaik...) and the code should be mostly pure.

VesaKarvonen:
Ok.  I can't promise anything, but I might take a look into implementing "async in parallel" support as I could use the functionality.

kurtschelfthout:
Promises not necessary, thanks for taking a look.

2015-01-27 11:05 GMT+00:00 Vesa Karvonen <notifications@github.com>:

VesaKarvonen:
I'm going to add some notes related to parallelizing FsCheck here.

The `Gen.variant` combinator does not appear to be safe for concurrent/parallel use: https://github.com/fsharp/FsCheck/blob/master/src/FsCheck/Gen.fs#L304

amazingant:
Sorry for a spurious comment, but +1 on this issue. I've found a decent handful of odd bugs by setting FsCheck to run my tests on the order of millions of times. Of course that means I've left my tests running overnight more than a few times. With parallelization, FsCheck would be an awesome use of any (currently) unused servers, because who doesn't love burning up 32 cores with F#?

amazingant:
And another random comment from the newcomer: For a quick hacky solution, I changed `Array.iter` to `Array.Parallel.iter` at [line 304 of Runner.fs][1]. I know this is not a good solution for running a single test with a large `MaxTest` value, but it's an easy change that gives me a very nice speed-up with my test suite.

[...]

Issue #310, opened 2016-09-23
Create ability to run a single test case.


bradphelan:
At the moment we can put in a seed to replicate an entire run of unit tests. However as our tests become more complicated they take longer to run. We have large numerical tests running on some big chunks of data. When we disover an edge case we need to debug it. However I can't see a way of only running the single test under failure.

ploeh:
When FsCheck finds input that falsifies your property, it reports the values in question in addition to the seed that caused the falsification. It both reports the original and shrunk values, if appropriate:

```fsharp
      Falsifiable, after 13 tests (11 shrinks) (StdGen (2016253613,296208608)):
      Original:
      UriScheme "kt@gY60TwEMr9c"
      Shrunk:
      UriScheme "kt@"
```

Is you property a `let` bound function, or perhaps a class method? If so, you can call it with the data reported by FsCheck, and debug it that way...

kurtschelfthout:
@ploeh True, and what I mostly end up doing. Also this gives you a regression unit test that you can (should?) keep around. 

But I still like the suggestion, and it should be relatively straightforward to implement.

kurtschelfthout:
Actually it is a bit tricky to implement, with the added constraint to keep backwards compatibility...

ploeh:
In the meantime, here's a trick that I just used, because I had a similar problem. Here's the output of a test that I ran:

```fsharp
------ Test started: Assembly: E247.Api.BuckarooPayment.ProtocolTests.dll ------
26 passed, 0 failed, 0 skipped, took 46,03 seconds (xUnit.net 2.1.0 build 3179).

------ Test started: Assembly: E247.Api.BuckarooPayment.UnitTests.dll ------

Test 'E247.Api.BuckarooPayment.BoundaryTests.validateCustomParameter returns correct result on present value' failed: FsCheck.Xunit.PropertyFailedException : 
Falsifiable, after 25 tests (4 shrinks) (StdGen (804078294,296222337)):
Original:
(NonEmptyString "P", ":wI")
[|{Name = "'W>c;r	
	----- Inner Stack Trace -----
	C:\Users\mark\Documents\24-7 Entertainment\Src\e247.api.buckaroopayment\BuckarooPayment.UnitTests\BoundaryTests.fs(72,0): at E247.Api.BuckarooPayment.BoundaryTests.validateCustomParameter returns correct result on present value@67.Invoke(CustomParameter[] customParameters)
[...]

PR #351, opened 2017-02-26
[WIP] Parallel generating and testing for #83

Rattenkrieg:
Do not merge.

Here is PoC for #83 
This PR introduces ability to run generation and checks in parallel through pluggable function [`parallelTest`](https://github.com/Rattenkrieg/FsCheck/blob/parallel-run-83/src/FsCheck/Runner.fs#L129)
 which is parallel version of [`test`](https://github.com/Rattenkrieg/FsCheck/blob/parallel-run-83/src/FsCheck/Runner.fs#L129)
Currently PLINQ is used as engine for parallel execution - choosen because it comes with framework and there is almost one-to-one translation of sequential routines. But any other options are welcomed.
Obviously there is now requirement on _Gens_ and properties to be thread-safe, so `StateMachine` and model-based tests should be runned in sequential mode.

Currently all tests are passing. Except `FsCheck.Experimental.Test.StateMachine` (surprisingly only one).
Although there is performance degradation due to scheduling overhead on default 100 passes, yet on 1000 passes running every test in parallel mode yields faster timings comparing to sequential

I will provide benchmarks using compute-bound _Gens_ and properties as well as properties with simulation of IO-bound work.

- [x] PoC for generating input and running properties in parallel
- [x] Ordering guarantees are consistent with sequential runner, we may want to provide ability to loosen it in favor of better performance through config
- [x] Find way to run tests without breaking incapsulation (maybe custom `IRunner`)
- [ ] Elaborate on whether to expose more tweaks through config (like degree of parallelism, order preservation etc)
- [x] True parallelism for asyncs/tasks
- [x] Benchmarks

### DIFF
--- a/docs/csharp/TestData.cs
+++ b/docs/csharp/TestData.cs
@@ -1,9 +1,5 @@
 ﻿using FsCheck;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace CSharp.DocSnippets {
 
--- a/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
+++ b/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
@@ -10,10 +10,18 @@ type NUnitTest() =
     [<Test>]
     member __.NormalTest() =
         ignore true
-
+        
     [<Property>]
     member __.RevUnit (x:char) =
         List.rev [x] = [x]
+
+    // Parallelism for PropertyAttribute not implemented for Nunit yet
+    // [<Property(Parallelism = 8)>]
+//    member __.Async (i:int) =
+//        async { 
+//            do! Async.Sleep 1500
+//            return true
+//        }
   
     [<Property>]
     member __.RevApp (x:string) xs =
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -13,4 +13,4 @@ nuget xunit.runner.console
 nuget xunit.runner.visualstudio version_in_path: true
 nuget FSharp.Formatting
 nuget Expecto
-nuget System.Threading
+nuget System.Threading.Thread
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -12,3 +12,5 @@ nuget xunit
 nuget xunit.runner.console
 nuget xunit.runner.visualstudio version_in_path: true
 nuget FSharp.Formatting
+nuget Expecto
+nuget System.Threading
--- a/src/FsCheck/Runner.fs
+++ b/src/FsCheck/Runner.fs
@@ -1,4 +1,4 @@
-namespace FsCheck
+﻿namespace FsCheck
 
 open System
 
@@ -33,6 +33,11 @@ type IRunner =
     ///Called whenever all tests are done, either True, False or Exhausted.
     abstract member OnFinished: string * TestResult -> unit
 
+type ParallelRunConfig =
+    { // For I/O bound work 1 would be fine, for cpu intensive tasks Environment.ProcessorCount appears to be fastest option
+      MaxDegreeOfParallelism : int 
+    }
+
 ///For configuring a run.
 [<NoEquality;NoComparison>]
 type Config = 
@@ -59,10 +64,12 @@ type Config =
       ///overridden)
       Arbitrary     : list<Type>
       ///A custom test runner, e.g. to integrate with a test framework like xUnit or NUnit. 
-      Runner        : IRunner }
+      Runner        : IRunner
+      ///If set, inputs for property generation and property evaluation will be runned in parallel. 
+      ParallelRunConfig : ParallelRunConfig option
+    }
 
 module Runner =
-
     open System.Collections.Generic
     open System.Reflection
 
@@ -79,50 +86,263 @@ module Runner =
         | Shrink of Result          //shrunk falsified result succesfully
         | NoShrink of Result        //could not falsify given result; so unsuccesful shrink.
         | EndShrink of Result       //gave up shrinking; (possibly) shrunk result is given
-
+        
+    [<NoEquality;NoComparison>]
+    type private OutcomeSeqOrFuture =
+        | Value of seq<TestStep>
+        | Future of Threading.Tasks.Task<seq<TestStep>>
  
-    let rec private shrinkResult (result:Result) (shrinks:IEnumerator<Rose<Result>>) =
-        seq { if (shrinks.MoveNext()) then
+    let rec private shrinkResultValue (result :Result) (shrinks :IEnumerator<Rose<Result>>) =
+        seq { 
+            if (shrinks.MoveNext ()) then
                 //result forced here
                 let (MkRose ((Lazy result'),shrinks')) = shrinks.Current
-                if result'.Outcome.Shrink then yield Shrink result'; yield! shrinkResult result' (shrinks'.GetEnumerator())
-                else yield NoShrink result'; yield! shrinkResult result shrinks
-              else
-                yield EndShrink result
+                if result'.Outcome.Shrink then yield Shrink result'; yield! shrinkResultValue result' (shrinks'.GetEnumerator())
+                else yield NoShrink result'; yield! shrinkResultValue result shrinks
+            else yield EndShrink result
         }
-        
-    let rec private test initSize resize rnd0 ((Gen eval) as gen) =
-        seq { // would like to get rid of this split,
-              // but the problem is DiscardException. If it is thrown,
-              // we also don't get the the updated Rnd back from eval.
-              let rnd1,rnd2 = Random.split rnd0
-              let newSize = resize initSize
-              //printfn "Before generate"
-              //result forced here!
-              let result, shrinks =
-                  try
-                    let (MkRose (Lazy result,shrinks)) = (eval (newSize |> round |> int) rnd2).Value
+ 
+    let rec private shrinkResultTask (result :Result) (shrinks :IEnumerator<Rose<ResultContainer>>) =
+        let goNext (s :seq<Rose<ResultContainer>>) r = 
+            async {
+                if r.Outcome.Shrink then 
+                    let! xs = shrinkResultTask r (s.GetEnumerator ()) |> Async.AwaitTask
+                    return seq { yield Shrink r; yield! xs }
+                else 
+                    let! xs = shrinkResultTask result shrinks |> Async.AwaitTask
+                    return seq { yield NoShrink r; yield! xs }
+            }
+        if (shrinks.MoveNext ()) then
+            //result forced here
+            let (MkRose ((Lazy result'),shrinks')) = shrinks.Current
+            match result' with
+            | ResultContainer.Value r -> 
+                goNext shrinks' r |> Async.StartAsTask
+            | ResultContainer.Future t ->
+                let rt = t |> Async.AwaitTask
+                async.Bind (rt, goNext shrinks') |> Async.StartAsTask
+        else EndShrink result |> Seq.singleton |> Threading.Tasks.Task.FromResult
+
+    let private shrinkResultTaskIter (result: Result) (shrinks :IEnumerator<Rose<ResultContainer>>) =
+        let xs :Collections.Generic.List<TestStep> = ResizeArray () 
+        let rec iter (result: Result) (shrinks :IEnumerator<Rose<ResultContainer>>) =
+            if (shrinks.MoveNext ()) then
+                //result forced here
+                let (MkRose ((Lazy result'), shrinks')) = shrinks.Current
+                match result' with
+                | ResultContainer.Value r -> 
+                    if r.Outcome.Shrink then 
+                        xs.Add <| Shrink r
+                        iter r (shrinks'.GetEnumerator ())
+                    else
+                        xs.Add <| NoShrink r
+                        iter result shrinks
+                | ResultContainer.Future t ->
+                    async {
+                        let! rt = t |> Async.AwaitTask
+                        if rt.Outcome.Shrink then 
+                            let! ys = shrinkResultTask rt (shrinks'.GetEnumerator ()) |> Async.AwaitTask
+                            return seq { yield! xs; yield Shrink rt; yield! ys }
+                        else 
+                            let! ys = shrinkResultTask result shrinks |> Async.AwaitTask
+                            return seq { yield! xs; yield NoShrink rt; yield! ys }
+                    } |> Async.StartAsTask |> OutcomeSeqOrFuture.Future
+            else xs.Add <| EndShrink result; xs :> seq<_> |> OutcomeSeqOrFuture.Value
+        iter result shrinks
+
+    let private test initSize resize rnd0 gen =
+        let rec test' initSize resize rnd0 ((Gen eval) as gen) =
+            seq { 
+                let rnd1,rnd2 = Random.split rnd0
+                let newSize = resize initSize
+                //result forced here!
+                let result, shrinks =
+                    try
+                        let (MkRose (Lazy result, shrinks)) = (eval (newSize |> round |> int) rnd2).Value
+                        result, shrinks
+                        //printfn "After generate"
+                        //problem: since result.Ok is no longer lazy, we only get the generated args _after_ the test is run
+                    with :? DiscardException -> 
+                        Res.rejectedV, Seq.empty
+
+                yield Generated result.Arguments
+
+                match result.Outcome with
+                    | Outcome.Rejected -> 
+                        yield Failed result 
+                    | Outcome.True -> 
+                        yield Passed result
+                    | o when o.Shrink -> 
+                        yield Falsified result
+                        yield! shrinkResultValue result (shrinks.GetEnumerator ())
+                    | _ ->
+                        yield Falsified result
+                        yield EndShrink result
+                yield! test' newSize resize rnd1 gen
+            }
+        //Since we're not runing test for parallel scenarios it's impossible to discover `ResultContainer.Future` inside `result`
+        let gen' = gen |> Gen.map (Rose.map (fun rc -> match rc with 
+                | ResultContainer.Value r -> r 
+                | ResultContainer.Future t -> t.Result))
+        test' initSize resize rnd0 gen'
+
+    let private outcomeSeq result (shrinks :seq<_>) = 
+        let g = Generated result.Arguments
+        match result.Outcome with
+        | Outcome.Rejected -> 
+            seq {yield g; yield Failed result} |> OutcomeSeqOrFuture.Value
+        | Outcome.True -> 
+            seq {yield g; yield Passed result} |> OutcomeSeqOrFuture.Value
+        | o when o.Shrink -> 
+            match shrinkResultTaskIter result (shrinks.GetEnumerator ()) with
+            | OutcomeSeqOrFuture.Value v -> seq {yield g; yield Falsified result; yield! v} |> OutcomeSeqOrFuture.Value
+            | OutcomeSeqOrFuture.Future t -> t.ContinueWith (fun (xs :Threading.Tasks.Task<seq<TestStep>>) -> 
+                seq {yield g; yield Falsified result; yield! xs.Result}) |> OutcomeSeqOrFuture.Future         
+        | _ ->
+            seq {yield g; yield Falsified result; yield EndShrink result} |> OutcomeSeqOrFuture.Value
+
+    let private testStep rnd (size :float) ((Gen eval) as gen) =
+            let result, shrinks =
+                try
+                    let (MkRose (Lazy result, shrinks)) = (eval (size |> round |> int) rnd).Value
                     result, shrinks
-                    //printfn "After generate"
-                    //problem: since result.Ok is no longer lazy, we only get the generated args _after_ the test is run
-                  with :? DiscardException -> 
+                with :? DiscardException ->
                     Res.rejected, Seq.empty
 
-              yield Generated result.Arguments
+            match result with
+            | ResultContainer.Value r -> outcomeSeq r shrinks
+            | ResultContainer.Future t -> t.ContinueWith (fun (x :Threading.Tasks.Task<Result>) -> 
+                match outcomeSeq x.Result shrinks with
+                | OutcomeSeqOrFuture.Value v -> Threading.Tasks.Task.FromResult v
+                | OutcomeSeqOrFuture.Future t -> t) |> Threading.Tasks.TaskExtensions.Unwrap |> OutcomeSeqOrFuture.Future
 
-              match result.Outcome with
-                | Outcome.Rejected -> 
-                    yield Failed result 
-                | Outcome.True -> 
-                    yield Passed result
-                | o when o.Shrink -> 
-                    yield Falsified result
-                    yield! shrinkResult result (shrinks.GetEnumerator())
-                | _ ->
-                    yield Falsified result
-                    yield EndShrink result
-              yield! test newSize resize rnd1 gen
-        }
+    let stepsSeq resize =
+        Seq.unfold (fun (initSize, rnd) ->
+            let rnd1, rnd2 = Random.split rnd
+            let newSize = resize initSize
+            Some ((rnd1, rnd2, newSize), (newSize, rnd1)))
+
+    let private outcomeSeqFutureCont (xs :Threading.Tasks.Task<seq<TestStep>>) (state :obj) =
+        match state with 
+        | :? (int * Threading.CancellationToken * array<seq<TestStep>> * int) as state ->
+            let (j, ct, results, iters) = state
+            if (not ct.IsCancellationRequested) && j < iters then
+                results.[j] <- xs.Result
+        | _ -> raise (System.ArgumentException ("state"))
+
+#if PCL
+    let private parallelTest config = test
+#else
+    let private tpWorkerFun (state :obj) =
+        match state with 
+        | :? ((Rnd * Rnd * float) array * (int ref) * int * Gen<Rose<ResultContainer>> * array<seq<TestStep>> * Threading.CancellationToken) as state ->
+            let (steps, i, iters, gen, results, ct) = state
+            let mutable j = 0
+            while (not ct.IsCancellationRequested) && j < iters do
+                j <- Threading.Interlocked.Increment (i) - 1
+                if j < iters then
+                    let _, rnd, size = steps.[j]
+                    let res = testStep rnd size gen
+                    match res with
+                    | OutcomeSeqOrFuture.Value xs -> results.[j] <- xs
+                    | OutcomeSeqOrFuture.Future ts -> 
+                        ts.ContinueWith (outcomeSeqFutureCont, (j, ct, results, iters)) |> ignore                    
+        | _ -> raise (System.ArgumentException ("state"))
+
+    ///Enumerates over `steps` seq publishing every item to `tpWorkerFun` via `ThreadPool.QueueUserWorkItem`
+    ///Waits for `tpWorkerFun` to populate `results` array
+    ///Order is preserved by sitting in busy loop & checking current entry in `results` 
+    ///    and yielding processor to other threads if cell haven't been populated yet
+    ///To be lean in allocated space first assumes that test would pass and creates `result` of size `maxTest`
+    ///    reallocates `result` to hold `maxFail` entries if test is falsified
+    ///Other strategies can be considered to trade between allocated memory, overall running time and amount of cross-process communication:
+    ///    - publish to `tpWorkerFun` more than one entry at a time
+    ///    - allocate `results` in lesser chunks in iterative manner
+    ///    - change busy loop with yelding to waiting on conditional variables (tested, leads to slower runing time)
+    type private ParSeqEnumerator (steps :IEnumerable<(Rnd * Rnd * float)>, maxTest, maxFail, maxDegreeOfParallelism, gen) =   
+        let size = maxTest + maxFail
+        let luckySeq = steps |> Seq.take maxTest
+        let unluckySeq = steps |> Seq.skip maxTest |> Seq.take maxFail
+        let mutable results = Array.zeroCreate<seq<TestStep>> maxTest
+        let mutable i = 0
+        let indexT = ref 0
+        let indexF = ref 0
+        let mutable current = Unchecked.defaultof<TestStep>
+        let mutable subE = Linq.Enumerable.Empty<TestStep>().GetEnumerator ()
+        let mutable cts = new Threading.CancellationTokenSource ()
+        let mutable started = false
+        let mutable firstRun = true
+        let run xs index =
+            let xs = Array.ofSeq xs
+            for i in 0..(Math.Min (Array.length xs, maxDegreeOfParallelism)) do
+                Threading.ThreadPool.QueueUserWorkItem (
+                    new Threading.WaitCallback (tpWorkerFun),
+                    (xs, index, Array.length xs, gen, results, cts.Token)) |> ignore
+        let moveNextInner () = 
+            if subE.MoveNext () then
+                current <- subE.Current; true  
+            else
+                subE.Dispose (); false    
+        let moveNextOuter () =
+            if i = maxTest && firstRun then
+                results <- Array.zeroCreate<seq<TestStep>> maxFail
+                i <- 0
+                run unluckySeq indexF
+                firstRun <- false
+            while isNull results.[i] do
+                Threading.Thread.Yield () |> ignore
+            subE <- results.[i].GetEnumerator ()
+            i <- i + 1
+        interface IEnumerator<TestStep> with
+            member __.MoveNext () = 
+                if not started then
+                    run luckySeq indexT; started <- true
+                let mutable running = true
+                let mutable moved = false
+                while running do
+                    if i >= size then
+                        running <- false
+                    else
+                        if moveNextInner () then
+                            running <- false
+                            moved <- true
+                        else
+                            moveNextOuter ()
+                moved
+            member __.Current :TestStep = current
+            member __.Current :obj = box current
+            member __.Reset () = 
+                cts.Cancel ()
+                cts.Dispose ()
+                cts <- new Threading.CancellationTokenSource ()
+                started <- false
+                firstRun <- true
+                results <- Array.zeroCreate<seq<TestStep>> maxTest
+                current <- Unchecked.defaultof<TestStep>
+                i <- 0
+                indexT := 0
+                indexF := 0
+                subE.Dispose ()
+        interface IDisposable with
+            member __.Dispose () = 
+                cts.Cancel ()
+                subE.Dispose ()
+                cts.Dispose ()
+
+
+    type private ParTestSeq (steps, maxTest, maxFail, maxDegreeOfParallelism, gen) =   
+        let enumerator () = new ParSeqEnumerator (steps, maxTest, maxFail, maxDegreeOfParallelism, gen)
+        interface IEnumerable<TestStep> with
+            member __.GetEnumerator () = enumerator () :> IEnumerator<TestStep>
+            member __.GetEnumerator () = enumerator () :> System.Collections.IEnumerator
+
+    let private parallelTest config initSize resize rnd0 gen =
+        let steps = stepsSeq resize (initSize, rnd0)
+        let pd = Option.fold (fun _ pc -> if pc.MaxDegreeOfParallelism <> -1 then pc.MaxDegreeOfParallelism else Environment.ProcessorCount) 1 config.ParallelRunConfig
+        let parSeq = ParTestSeq (steps, config.MaxTest, config.MaxFail, pd, gen)
+        parSeq :> seq<_>
+#endif
+
 
     let private testsDone config testStep origArgs ntest nshrinks usedSeed stamps  =
         let entry (n,xs) = (100 * n / ntest),xs
@@ -144,19 +364,24 @@ module Runner =
                 | _ -> failwith "Test ended prematurely"
         config.Runner.OnFinished(config.Name,testResult)
 
+
     let private runner config prop = 
         let testNb = ref 0
         let failedNb = ref 0
         let shrinkNb = ref 0
         let tryShrinkNb = ref 0
         let origArgs = ref []
-        let lastStep = ref (Failed Res.rejected)
+        let lastStep = ref (Failed Res.rejectedV)
         let seed = match config.Replay with None -> Random.create() | Some s -> s
         let increaseSizeStep = float (config.EndSize - config.StartSize) / float config.MaxTest
-        test (float config.StartSize) ((+) increaseSizeStep) seed (property prop |> Property.GetGen) 
+        let testSeq = 
+            if config.ParallelRunConfig.IsSome then 
+                parallelTest config 
+             else 
+                test
+        testSeq (float config.StartSize) ((+) increaseSizeStep) seed (property prop |> Property.GetGen)
         |> Common.takeWhilePlusLast (fun step ->
             lastStep := step
-            //printfn "%A" step
             match step with
                 | Generated args -> config.Runner.OnArguments(!testNb, args, config.Every); true
                 | Passed _ -> testNb := !testNb + 1; !testNb <> config.MaxTest //stop if we have enough tests
@@ -335,7 +560,8 @@ type Config with
               Every         = fun _ _ -> String.Empty
               EveryShrink   = fun _ -> String.Empty
               Arbitrary     = []
-              Runner        = consoleRunner } 
+              Runner        = consoleRunner
+              ParallelRunConfig = None } 
 
     ///The verbose configuration prints each generated argument.
     static member Verbose = 
--- a/src/FsCheck/Runner.fs
+++ b/src/FsCheck/Runner.fs
@@ -290,7 +290,11 @@ module Runner =
                 run unluckySeq indexF
                 firstRun <- false
             while isNull results.[i] do
+#if NETSTANDARD1_6
+                Threading.Thread.Sleep (1) |> ignore
+#else
                 Threading.Thread.Yield () |> ignore
+#endif
             subE <- results.[i].GetEnumerator ()
             i <- i + 1
         interface IEnumerator<TestStep> with
--- a/src/FsCheck/Testable.fs
+++ b/src/FsCheck/Testable.fs
@@ -1,5 +1,7 @@
 ﻿namespace FsCheck
 
+open System
+
 [<NoComparison; RequireQualifiedAccess>]
 type Outcome = 
     | Timeout of int
@@ -19,7 +21,7 @@ type Result =
         Arguments   : list<obj> } 
     ///Returns a new result that is Succeeded if and only if both this
     ///and the given Result are Succeeded.
-    static member (&&&) (l,r) = 
+    static member resAnd l r = 
         //printfn "And of l %A and r %A" l.Outcome r.Outcome
         match (l.Outcome,r.Outcome) with
         | (Outcome.Exception _,_) -> l //here a potential exception in r is thrown away...
@@ -31,7 +33,7 @@ type Result =
         | (_,Outcome.True) -> l
         | (Outcome.True,_) -> r
         | (Outcome.Rejected,Outcome.Rejected) -> l //or r, whatever
-    static member (|||) (l,r) =
+    static member resOr l r =
         match (l.Outcome, r.Outcome) with
         | (Outcome.Exception _,_) -> r
         | (_,Outcome.Exception _) -> l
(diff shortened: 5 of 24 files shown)


kurtschelfthout:
Nice. It looks like the .NET Core build fails; not entirely clear what is up with that. Perhaps just need to add the PLINQ assemblies? If not supported on NetStandard 1.6 may  have to conditionally compile...

src/FsCheck/Runner.fs
```diff
@@ -120,6 +122,42 @@ module Runner =
...
               yield! test newSize resize rnd1 gen
         }
+        
+#if PCL // sorry windows phone no parallelism for you
+    let private parallelTest limit = test
+#else
+    let internal parallelTest limit initSize resize rnd0 gen =
+        (initSize, rnd0)
+        |> Seq.unfold (fun (initSize, rnd) ->
+            let rnd1, rnd2 = Random.split rnd
```
kurtschelfthout: Note here that really, Random.split is intended to start a new "fork" of random numbers ( a new parallel thread/task) while the `next` methods is intended to get the next seed in the current fork. I still need to update this in the sequential runner too on this fork, but I would expect for the sequential runner that `split` is not called where it is now, and we use `next` instead. For the parallel case ideally a mix of the two is needed. This is necessary to have good randomness properties.

Rattenkrieg: If I understand you properly then you are about to rewrite sequential `test` like this:

```fsharp
   let rec private test initSize resize rnd0 gen =
        seq { let rnd1 = rnd0.Next ()
              let newSize = resize initSize
              let result, shrinks =
                  try
                    let (MkRose (Lazy result,shrinks)) = Gen.eval (newSize |> round |> int) rnd1 gen
                    result, shrinks
                  with :? DiscardException -> 
                    Res.rejected, Seq.empty

              // yield TestStep / shrinking
              yield! test newSize resize rnd1 gen
        }
```

Now same `rnd1` is used for `Gen.eval` and param to next call of `test`

My concern is to keep results of parallel execution consistent with results of sequential algorithm. Otherwise user will have hard times if he first found falsifiable seed in _par_ and then switching to _seq_ with same seed for debugging.

kurtschelfthout: > If I understand you properly then you are about to rewrite sequential test like this:

Yes, that's what I had in mind.

>My concern is to keep results of parallel execution consistent with results of sequential algorithm. Otherwise user will have hard times if he first found falsifiable seed in par and then switching to seq with same seed for debugging.

That is a valid concern, I think largely addressed by #310 with the ability to run a single test case with a given seed and size (i.e. one case, trivially sequential).

The advantage of using `split` for parallel execution is that you could parallelize the whole thing, whereas now you basically have a purely sequential part first to generate the random values and size. Perhaps this is unlikely to make a big difference in practice - generating random values should be pretty fast. I also don't know if it would work well with the parallelism API. But I thought I'd mention it in case you are interested to explore the option or the benchmarks don't come out looking totally awesome ;)

For this PR, I'd be  totally fine with keeping things as is (i.e. using `split`) and we can look at using `next` instead later. If we keep the seeds as pre-generated in the parallel run, switching `split` to `next` should be trivial to do later anyway.

Rattenkrieg: Regarding benchmarks. Well... currently I've only ran tight compute-bound benchmarks, and plinq is slightly behind(!) of sequential runner with about 10-20x memory consumption(!!!). So I've implemented another parallel version using threadpool which is 1.5-2 times faster than sequential (on my 4-core laptop). But anyway I expected speedups to be somewhere close to linear. I need little more time to craft io-bound (maybe artificially via `Task.Delay`) and will provide full report then.

kurtschelfthout: I see. Also note that in some (most?) cases, the size of the generated value influences both the time it takes to generate it (i.e. an n-ary tree of depth 50 takes longer to generate than one of depth 1) and likely the time to run the test as well. It may be necessary to have a fixed (configurable) degree of parallelism, and then assign tests to threads in a likely better balanced way. E.g. say for 2 threads, instead of:
- thread 1: size 1 -> 25
- thread 2: size 25 -> 50

You'd do something like assign all the even sizes to thread 1 and the odd sizes to thread 2.

It's actually possible that on some compute-bound benchmarks plinq would perform better there, as I understand it has fancy work-stealing algorithms etc.

Rattenkrieg: Actually I'm crafting next version of parallel runner (third one) with algorithm behind like you described, running on bare threads with no contention on producer seq. Degree of parallelism currently supported in config (have not pushed yet, code is dirty right now :rage1:) but seems it has almost no effect on plinq. Lot of performance ruined by requirement to preserve ordering among produced results though plinq docs says there will be little degradation. My algorithm isn't affected by this since it preallocates result array upfront and each thread has it own index per iteration.

Rattenkrieg:
As PLINQ appeared really slow, I have implemented custom parallel runner. Here comes several benchmarks for it


<details><summary>FsCheckParallelBenchSort</summary><p>


```
                  Method |       Jit | Platform | Server |      Mean |   StdErr |   StdDev |
------------------------ |---------- |--------- |------- |---------- |--------- |--------- |
                 SeqSort | LegacyJit |      X64 |  False | 27.0369 s | 0.0534 s | 0.1068 s |
 DedicatedParallelSort16 | LegacyJit |      X64 |  False | 16.6396 s | 0.1451 s | 0.2901 s |
  DedicatedParallelSort8 | LegacyJit |      X64 |  False | 15.9758 s | 0.1022 s | 0.2043 s |
  DedicatedParallelSort4 | LegacyJit |      X64 |  False | 15.2732 s | 0.0702 s | 0.1403 s |
                 SeqSort | LegacyJit |      X64 |   True | 24.9697 s | 0.0314 s | 0.0629 s |
 DedicatedParallelSort16 | LegacyJit |      X64 |   True |  7.6853 s | 0.0656 s | 0.1311 s |
  DedicatedParallelSort8 | LegacyJit |      X64 |   True |  7.7246 s | 0.0452 s | 0.0903 s |
  DedicatedParallelSort4 | LegacyJit |      X64 |   True |  9.9426 s | 0.0359 s | 0.0717 s |
                 SeqSort | LegacyJit |      X86 |  False | 22.0300 s | 0.0458 s | 0.0916 s |
 DedicatedParallelSort16 | LegacyJit |      X86 |  False | 15.8121 s | 0.1453 s | 0.2906 s |
  DedicatedParallelSort8 | LegacyJit |      X86 |  False | 15.4747 s | 0.0168 s | 0.0336 s |
  DedicatedParallelSort4 | LegacyJit |      X86 |  False | 13.9290 s | 0.0818 s | 0.1635 s |
                 SeqSort | LegacyJit |      X86 |   True | 19.4483 s | 0.0333 s | 0.0665 s |
 DedicatedParallelSort16 | LegacyJit |      X86 |   True |  5.8886 s | 0.0528 s | 0.1056 s |
  DedicatedParallelSort8 | LegacyJit |      X86 |   True |  5.9589 s | 0.0770 s | 0.1540 s |
  DedicatedParallelSort4 | LegacyJit |      X86 |   True |  8.1849 s | 0.0438 s | 0.0875 s |
                 SeqSort |    RyuJit |      X64 |  False | 27.2159 s | 0.0359 s | 0.0719 s |
 DedicatedParallelSort16 |    RyuJit |      X64 |  False | 18.7917 s | 0.3023 s | 0.6046 s |
  DedicatedParallelSort8 |    RyuJit |      X64 |  False | 18.0660 s | 0.0199 s | 0.0398 s |
  DedicatedParallelSort4 |    RyuJit |      X64 |  False | 15.8674 s | 0.0463 s | 0.0926 s |
                 SeqSort |    RyuJit |      X64 |   True | 25.1002 s | 0.0619 s | 0.1238 s |
 DedicatedParallelSort16 |    RyuJit |      X64 |   True |  7.4592 s | 0.0985 s | 0.1971 s |
  DedicatedParallelSort8 |    RyuJit |      X64 |   True |  7.5945 s | 0.0608 s | 0.1216 s |
  DedicatedParallelSort4 |    RyuJit |      X64 |   True |  9.8323 s | 0.0243 s | 0.0485 s |
                 SeqSort |    RyuJit |      X86 |  False | 21.9957 s | 0.0206 s | 0.0411 s |
 DedicatedParallelSort16 |    RyuJit |      X86 |  False | 15.7843 s | 0.0493 s | 0.0986 s |
  DedicatedParallelSort8 |    RyuJit |      X86 |  False | 15.4891 s | 0.0257 s | 0.0515 s |
  DedicatedParallelSort4 |    RyuJit |      X86 |  False | 13.9200 s | 0.0610 s | 0.1221 s |
                 SeqSort |    RyuJit |      X86 |   True | 19.3723 s | 0.0131 s | 0.0261 s |
 DedicatedParallelSort16 |    RyuJit |      X86 |   True |  5.9516 s | 0.0567 s | 0.1135 s |
  DedicatedParallelSort8 |    RyuJit |      X86 |   True |  5.9669 s | 0.0356 s | 0.0711 s |
  DedicatedParallelSort4 |    RyuJit |      X86 |   True |  7.9802 s | 0.0654 s | 0.1309 s |
```
</p></details>


<details><summary>FsCheckParallelBenchPrimes</summary><p>


```
                    Method |       Jit | Platform | Server |      Mean |   StdErr |   StdDev |
-------------------------- |---------- |--------- |------- |---------- |--------- |--------- |
                 SeqPrimes | LegacyJit |      X64 |  False | 43.9916 s | 0.1571 s | 0.3142 s |
 DedicatedParallelPrimes16 | LegacyJit |      X64 |  False | 16.9838 s | 0.0778 s | 0.1557 s |
  DedicatedParallelPrimes8 | LegacyJit |      X64 |  False | 16.9262 s | 0.0760 s | 0.1521 s |
  DedicatedParallelPrimes4 | LegacyJit |      X64 |  False | 19.3157 s | 0.0493 s | 0.0986 s |
                 SeqPrimes | LegacyJit |      X64 |   True | 42.5074 s | 0.1949 s | 0.3898 s |
 DedicatedParallelPrimes16 | LegacyJit |      X64 |   True | 13.5379 s | 0.1451 s | 0.2903 s |
  DedicatedParallelPrimes8 | LegacyJit |      X64 |   True | 13.6245 s | 0.1135 s | 0.2269 s |
  DedicatedParallelPrimes4 | LegacyJit |      X64 |   True | 16.3157 s | 0.1798 s | 0.3595 s |
                 SeqPrimes | LegacyJit |      X86 |  False | 19.5355 s | 0.0493 s | 0.0987 s |
 DedicatedParallelPrimes16 | LegacyJit |      X86 |  False |  9.9948 s | 0.0498 s | 0.0996 s |
  DedicatedParallelPrimes8 | LegacyJit |      X86 |  False | 10.0483 s | 0.0190 s | 0.0380 s |
  DedicatedParallelPrimes4 | LegacyJit |      X86 |  False | 10.6440 s | 0.0929 s | 0.1857 s |
                 SeqPrimes | LegacyJit |      X86 |   True | 16.7634 s | 0.0200 s | 0.0400 s |
 DedicatedParallelPrimes16 | LegacyJit |      X86 |   True |  6.2209 s | 0.0315 s | 0.0630 s |
  DedicatedParallelPrimes8 | LegacyJit |      X86 |   True |  6.1674 s | 0.0252 s | 0.0504 s |
  DedicatedParallelPrimes4 | LegacyJit |      X86 |   True |  7.0708 s | 0.0270 s | 0.0541 s |
                 SeqPrimes |    RyuJit |      X64 |  False | 39.4370 s | 0.2967 s | 0.5935 s |
 DedicatedParallelPrimes16 |    RyuJit |      X64 |  False | 15.6690 s | 0.0742 s | 0.1483 s |
  DedicatedParallelPrimes8 |    RyuJit |      X64 |  False | 15.5193 s | 0.0410 s | 0.0820 s |
  DedicatedParallelPrimes4 |    RyuJit |      X64 |  False | 18.9811 s | 0.7588 s | 1.5176 s |
                 SeqPrimes |    RyuJit |      X64 |   True | 38.2184 s | 0.0797 s | 0.1595 s |
 DedicatedParallelPrimes16 |    RyuJit |      X64 |   True | 12.1975 s | 0.0592 s | 0.1184 s |
  DedicatedParallelPrimes8 |    RyuJit |      X64 |   True | 12.2878 s | 0.1662 s | 0.3324 s |
  DedicatedParallelPrimes4 |    RyuJit |      X64 |   True | 14.6240 s | 0.1088 s | 0.2176 s |
                 SeqPrimes |    RyuJit |      X86 |  False | 19.5087 s | 0.0710 s | 0.1421 s |
 DedicatedParallelPrimes16 |    RyuJit |      X86 |  False | 10.0410 s | 0.0308 s | 0.0616 s |
  DedicatedParallelPrimes8 |    RyuJit |      X86 |  False | 10.0532 s | 0.1019 s | 0.2038 s |
  DedicatedParallelPrimes4 |    RyuJit |      X86 |  False | 10.6213 s | 0.0405 s | 0.0811 s |
                 SeqPrimes |    RyuJit |      X86 |   True | 16.7677 s | 0.0610 s | 0.1221 s |
 DedicatedParallelPrimes16 |    RyuJit |      X86 |   True |  6.2381 s | 0.0454 s | 0.0909 s |
  DedicatedParallelPrimes8 |    RyuJit |      X86 |   True |  6.1650 s | 0.0314 s | 0.0628 s |
  DedicatedParallelPrimes4 |    RyuJit |      X86 |   True |  7.0610 s | 0.0464 s | 0.0928 s |
```
</p></details>


<details><summary>FsCheckParallelBench</summary><p>

this one based on my actual test where I'm transforming 2k jsons to csv. Lot of allocations on _Gen_ side, almost no allocations on algorithm under test side.
```
                    Method |       Jit | Platform | Server |      Mean |   StdErr |   StdDev |
-------------------------- |---------- |--------- |------- |---------- |--------- |--------- |
  DedicatedParallelCount16 | LegacyJit |      X64 |  False | 16.3143 s | 0.0581 s | 0.1162 s |
   DedicatedParallelCount8 | LegacyJit |      X64 |  False | 16.1575 s | 0.1168 s | 0.2335 s |
   DedicatedParallelCount4 | LegacyJit |      X64 |  False | 18.4367 s | 0.0654 s | 0.1308 s |
                  SeqCount | LegacyJit |      X64 |  False | 42.5618 s | 0.1985 s | 0.3970 s |
  DedicatedParallelWidth16 | LegacyJit |      X64 |  False | 16.2795 s | 0.0945 s | 0.1890 s |
   DedicatedParallelWidth8 | LegacyJit |      X64 |  False | 16.2165 s | 0.0438 s | 0.0877 s |
   DedicatedParallelWidth4 | LegacyJit |      X64 |  False | 18.6480 s | 0.0795 s | 0.1589 s |
                  SeqWidth | LegacyJit |      X64 |  False | 42.6715 s | 0.1858 s | 0.3716 s |
 DedicatedParallelEquals16 | LegacyJit |      X64 |  False | 16.9074 s | 0.0974 s | 0.1948 s |
  DedicatedParallelEquals8 | LegacyJit |      X64 |  False | 16.7771 s | 0.0611 s | 0.1222 s |
  DedicatedParallelEquals4 | LegacyJit |      X64 |  False | 19.2614 s | 0.1081 s | 0.2161 s |
                 SeqEquals | LegacyJit |      X64 |  False | 44.1847 s | 0.4897 s | 0.9794 s |
  DedicatedParallelCount16 | LegacyJit |      X64 |   True | 13.1924 s | 0.1481 s | 0.2962 s |
   DedicatedParallelCount8 | LegacyJit |      X64 |   True | 13.0821 s | 0.1241 s | 0.2483 s |
   DedicatedParallelCount4 | LegacyJit |      X64 |   True | 15.5106 s | 0.2189 s | 0.4378 s |
                  SeqCount | LegacyJit |      X64 |   True | 40.9960 s | 0.2375 s | 0.4750 s |
  DedicatedParallelWidth16 | LegacyJit |      X64 |   True | 12.9859 s | 0.0909 s | 0.1818 s |
   DedicatedParallelWidth8 | LegacyJit |      X64 |   True | 12.9674 s | 0.0824 s | 0.1647 s |
   DedicatedParallelWidth4 | LegacyJit |      X64 |   True | 15.3704 s | 0.1812 s | 0.3625 s |
                  SeqWidth | LegacyJit |      X64 |   True | 40.7517 s | 0.1562 s | 0.3123 s |
 DedicatedParallelEquals16 | LegacyJit |      X64 |   True | 13.8073 s | 0.0868 s | 0.1737 s |
  DedicatedParallelEquals8 | LegacyJit |      X64 |   True | 13.6996 s | 0.1385 s | 0.2769 s |
  DedicatedParallelEquals4 | LegacyJit |      X64 |   True | 16.1911 s | 0.0827 s | 0.1654 s |
                 SeqEquals | LegacyJit |      X64 |   True | 42.8135 s | 0.1975 s | 0.3950 s |
  DedicatedParallelCount16 | LegacyJit |      X86 |  False |  9.2344 s | 0.0373 s | 0.0745 s |
   DedicatedParallelCount8 | LegacyJit |      X86 |  False |  9.3421 s | 0.0548 s | 0.1096 s |
   DedicatedParallelCount4 | LegacyJit |      X86 |  False |  9.7986 s | 0.0895 s | 0.1789 s |
                  SeqCount | LegacyJit |      X86 |  False | 18.0496 s | 0.0780 s | 0.1561 s |
  DedicatedParallelWidth16 | LegacyJit |      X86 |  False |  9.3402 s | 0.0510 s | 0.1021 s |
   DedicatedParallelWidth8 | LegacyJit |      X86 |  False |  9.3471 s | 0.0429 s | 0.0857 s |
   DedicatedParallelWidth4 | LegacyJit |      X86 |  False | 10.1183 s | 0.0770 s | 0.1541 s |
                  SeqWidth | LegacyJit |      X86 |  False | 18.1846 s | 0.1462 s | 0.2924 s |
 DedicatedParallelEquals16 | LegacyJit |      X86 |  False | 10.0501 s | 0.0750 s | 0.1500 s |
  DedicatedParallelEquals8 | LegacyJit |      X86 |  False | 10.0009 s | 0.0327 s | 0.0654 s |
  DedicatedParallelEquals4 | LegacyJit |      X86 |  False | 10.6096 s | 0.0183 s | 0.0366 s |
                 SeqEquals | LegacyJit |      X86 |  False | 19.6891 s | 0.0886 s | 0.1772 s |
  DedicatedParallelCount16 | LegacyJit |      X86 |   True |  5.6234 s | 0.0726 s | 0.1452 s |
   DedicatedParallelCount8 | LegacyJit |      X86 |   True |  5.6097 s | 0.0370 s | 0.0741 s |
   DedicatedParallelCount4 | LegacyJit |      X86 |   True |  6.4165 s | 0.0253 s | 0.0505 s |
                  SeqCount | LegacyJit |      X86 |   True | 14.9982 s | 0.0470 s | 0.0940 s |
  DedicatedParallelWidth16 | LegacyJit |      X86 |   True |  5.6435 s | 0.0370 s | 0.0739 s |
   DedicatedParallelWidth8 | LegacyJit |      X86 |   True |  5.6340 s | 0.0424 s | 0.0849 s |
   DedicatedParallelWidth4 | LegacyJit |      X86 |   True |  6.4176 s | 0.0462 s | 0.0923 s |
                  SeqWidth | LegacyJit |      X86 |   True | 15.2575 s | 0.0180 s | 0.0359 s |
 DedicatedParallelEquals16 | LegacyJit |      X86 |   True |  6.1345 s | 0.0410 s | 0.0821 s |
  DedicatedParallelEquals8 | LegacyJit |      X86 |   True |  6.1778 s | 0.0452 s | 0.0905 s |
  DedicatedParallelEquals4 | LegacyJit |      X86 |   True |  7.0763 s | 0.0592 s | 0.1184 s |
                 SeqEquals | LegacyJit |      X86 |   True | 16.8870 s | 0.0817 s | 0.1634 s |
  DedicatedParallelCount16 |    RyuJit |      X64 |  False | 14.7642 s | 0.0277 s | 0.0554 s |
   DedicatedParallelCount8 |    RyuJit |      X64 |  False | 14.7435 s | 0.0378 s | 0.0757 s |
   DedicatedParallelCount4 |    RyuJit |      X64 |  False | 16.9032 s | 0.0580 s | 0.1159 s |
                  SeqCount |    RyuJit |      X64 |  False | 37.9066 s | 0.2136 s | 0.4272 s |
  DedicatedParallelWidth16 |    RyuJit |      X64 |  False | 14.9439 s | 0.1316 s | 0.2632 s |
   DedicatedParallelWidth8 |    RyuJit |      X64 |  False | 14.9065 s | 0.0810 s | 0.1620 s |
   DedicatedParallelWidth4 |    RyuJit |      X64 |  False | 16.8898 s | 0.0212 s | 0.0423 s |
                  SeqWidth |    RyuJit |      X64 |  False | 38.3218 s | 0.2762 s | 0.5525 s |
 DedicatedParallelEquals16 |    RyuJit |      X64 |  False | 15.7063 s | 0.0854 s | 0.1707 s |
  DedicatedParallelEquals8 |    RyuJit |      X64 |  False | 15.5631 s | 0.0492 s | 0.0984 s |
  DedicatedParallelEquals4 |    RyuJit |      X64 |  False | 17.7210 s | 0.0682 s | 0.1365 s |
                 SeqEquals |    RyuJit |      X64 |  False | 39.5402 s | 0.1681 s | 0.3362 s |
  DedicatedParallelCount16 |    RyuJit |      X64 |   True | 11.6130 s | 0.1430 s | 0.2860 s |
   DedicatedParallelCount8 |    RyuJit |      X64 |   True | 11.6754 s | 0.1647 s | 0.3294 s |
   DedicatedParallelCount4 |    RyuJit |      X64 |   True | 13.9093 s | 0.1461 s | 0.2921 s |
                  SeqCount |    RyuJit |      X64 |   True | 36.5093 s | 0.0804 s | 0.1608 s |
  DedicatedParallelWidth16 |    RyuJit |      X64 |   True | 11.6031 s | 0.1308 s | 0.2617 s |
   DedicatedParallelWidth8 |    RyuJit |      X64 |   True | 11.5663 s | 0.1125 s | 0.2250 s |
   DedicatedParallelWidth4 |    RyuJit |      X64 |   True | 13.8691 s | 0.2179 s | 0.4357 s |
                  SeqWidth |    RyuJit |      X64 |   True | 36.6191 s | 0.1992 s | 0.3983 s |
 DedicatedParallelEquals16 |    RyuJit |      X64 |   True | 12.2522 s | 0.0281 s | 0.0562 s |
  DedicatedParallelEquals8 |    RyuJit |      X64 |   True | 12.3276 s | 0.0882 s | 0.1764 s |
  DedicatedParallelEquals4 |    RyuJit |      X64 |   True | 14.5557 s | 0.1324 s | 0.2649 s |
                 SeqEquals |    RyuJit |      X64 |   True | 38.0303 s | 0.1818 s | 0.3637 s |
  DedicatedParallelCount16 |    RyuJit |      X86 |  False |  9.2771 s | 0.0606 s | 0.1212 s |
   DedicatedParallelCount8 |    RyuJit |      X86 |  False |  9.2733 s | 0.0493 s | 0.0985 s |
   DedicatedParallelCount4 |    RyuJit |      X86 |  False |  9.7757 s | 0.0976 s | 0.1951 s |
                  SeqCount |    RyuJit |      X86 |  False | 17.8693 s | 0.1084 s | 0.2167 s |
  DedicatedParallelWidth16 |    RyuJit |      X86 |  False |  9.2663 s | 0.0981 s | 0.1963 s |
   DedicatedParallelWidth8 |    RyuJit |      X86 |  False |  9.3708 s | 0.0598 s | 0.1196 s |
   DedicatedParallelWidth4 |    RyuJit |      X86 |  False |  9.8846 s | 0.0628 s | 0.1256 s |
                  SeqWidth |    RyuJit |      X86 |  False | 18.1215 s | 0.0893 s | 0.1787 s |
 DedicatedParallelEquals16 |    RyuJit |      X86 |  False | 10.1015 s | 0.0296 s | 0.0591 s |
  DedicatedParallelEquals8 |    RyuJit |      X86 |  False | 10.0373 s | 0.0645 s | 0.1290 s |
  DedicatedParallelEquals4 |    RyuJit |      X86 |  False | 10.5324 s | 0.0886 s | 0.1771 s |
                 SeqEquals |    RyuJit |      X86 |  False | 19.5843 s | 0.1052 s | 0.2105 s |
  DedicatedParallelCount16 |    RyuJit |      X86 |   True |  5.5009 s | 0.0117 s | 0.0235 s |
   DedicatedParallelCount8 |    RyuJit |      X86 |   True |  5.5607 s | 0.0252 s | 0.0504 s |
   DedicatedParallelCount4 |    RyuJit |      X86 |   True |  6.3665 s | 0.0335 s | 0.0670 s |
                  SeqCount |    RyuJit |      X86 |   True | 15.0297 s | 0.0665 s | 0.1329 s |
  DedicatedParallelWidth16 |    RyuJit |      X86 |   True |  5.6187 s | 0.0253 s | 0.0507 s |
   DedicatedParallelWidth8 |    RyuJit |      X86 |   True |  5.6014 s | 0.0185 s | 0.0370 s |
   DedicatedParallelWidth4 |    RyuJit |      X86 |   True |  6.4729 s | 0.0150 s | 0.0300 s |
                  SeqWidth |    RyuJit |      X86 |   True | 15.2337 s | 0.0720 s | 0.1440 s |
 DedicatedParallelEquals16 |    RyuJit |      X86 |   True |  6.1754 s | 0.0319 s | 0.0638 s |
  DedicatedParallelEquals8 |    RyuJit |      X86 |   True |  6.1555 s | 0.0331 s | 0.0662 s |
  DedicatedParallelEquals4 |    RyuJit |      X86 |   True |  7.1158 s | 0.0314 s | 0.0629 s |
                 SeqEquals |    RyuJit |      X86 |   True | 16.8272 s | 0.0830 s | 0.1660 s |
```
</p></details>


Due to weird issue on BDN side (I think this one dotnet/BenchmarkDotNet#377) I was unable to produce report for `FsCheckParallelBenchIo` but it scales in precisely linear fashion with `MaxDegreeOfParallelism = Environment.ProcessorCount`. However doubling `MaxDegreeOfParallelism` yields worse timings. Probably we should not run async work on current parallel-test-runner (threadpool's threads) rather it will be convenient to treat async tests as first-class citizens (like proposed in #167) and leverage `TaskScheduler`. That way async tests can be ran in parallel with more throughput than `Environment.ProcessorCount`. I'm already using hack for it in production, see [this gist](https://gist.github.com/Rattenkrieg/b50e6b38e37c749a0a0c4f4b2bdaefbd). Maybe @polytypic and @amazingant would share their scenarios.

Regarding benchmarks I have observed lot of GC interventions preventing runner to scale lineary. Concurrent-Server GC yields higher scalability ratio (with more memory consumption) though I don't expect anyone would run tests on such preset.

src/FsCheck/Runner.fs
```diff
@@ -71,7 +76,7 @@ module Runner =
     open TypeClass
    
     [<NoEquality;NoComparison>]
-    type private TestStep = 
+    type TestStep =
```
kurtschelfthout: Why not private?

src/FsCheck/Runner.fs
```diff
@@ -59,10 +63,11 @@ type Config =
       ///overridden)
       Arbitrary     : list<Type>
       ///A custom test runner, e.g. to integrate with a test framework like xUnit or NUnit. 
-      Runner        : IRunner }
+      Runner        : IRunner
+      ///If set, inputs for property generation and property evaluation will be runned in parallel. 
+      ParallelRunConfig : ParallelRunConfig option}
```
kurtschelfthout: Do you have any other options in mind? If not would prefer to just put the one field into the Config record.

kurtschelfthout:
This is starting to look really good.

I see you added projects with dependencies - please use paket to add all such dependencies (i.e. change paket.dependencies and add project specific dependencies files, instead of packages.config). This may be why the build failed.

src/FsCheck/Runner.fs
Rattenkrieg: I have forgot to cleanup, will fix it in next commit

src/FsCheck/Runner.fs
Rattenkrieg: First I've added `preserveOrdering :bool` and `algorithm :int`. But benchmarks eliminated all algorithms except current one and preserving ordering was suitable only for plinq. If you have no other options in mind I can leave just bool flag here.

Rattenkrieg:
Those projects are only for demonstration what scenarios I've used for benchmarks. I'm about to remove it (or maybe move to separate repo if anyone would found it usefull) after we'll elaborate api.

kurtschelfthout:
> But if you want to see builds passing, I'm fine to fix it now.

Not at all, I was just commenting because I thought perhaps you were wondering if anyone was even paying attention :) Follow whatever time frame you like and work in whatever way you like.

I guess let me know when you'd like some comments.

Rattenkrieg:
Finally I've managed to hack `asyncs` and `tasks` so we can run such cases with mad degree of parallelism. Here are results for [fetching google's front page](https://github.com/Rattenkrieg/fscheckbenchmarks/blob/master/src/ParallelBenchmarks/Properties.fs#L148):

<details><summary>FsCheckParallelBenchSort</summary><p>

```
                Method |       Jit | Platform | Server |        Mean |        Error |       StdDev |
---------------------- |---------- |--------- |------- |------------:|-------------:|-------------:|
                 SeqIo | LegacyJit |      X64 |  False | 49,148.5 ms | 30,830.22 ms | 4,779.565 ms |
 DedicatedParallelIo16 | LegacyJit |      X64 |  False |  1,271.1 ms |    524.89 ms |    81.373 ms |
  DedicatedParallelIo8 | LegacyJit |      X64 |  False |  1,362.3 ms |    917.50 ms |   142.238 ms |
  DedicatedParallelIo4 | LegacyJit |      X64 |  False |  1,337.6 ms |     46.16 ms |     7.156 ms |
                 SeqIo | LegacyJit |      X64 |   True | 50,260.3 ms | 41,653.92 ms | 6,457.548 ms |
 DedicatedParallelIo16 | LegacyJit |      X64 |   True |  1,058.6 ms |    118.64 ms |    18.392 ms |
  DedicatedParallelIo8 | LegacyJit |      X64 |   True |  1,106.8 ms |    548.11 ms |    84.973 ms |
  DedicatedParallelIo4 | LegacyJit |      X64 |   True |  1,145.9 ms |    215.42 ms |    33.396 ms |
                 SeqIo | LegacyJit |      X86 |  False | 49,013.3 ms | 25,453.79 ms | 3,946.065 ms |
 DedicatedParallelIo16 | LegacyJit |      X86 |  False |  1,046.5 ms |    164.79 ms |    25.546 ms |
  DedicatedParallelIo8 | LegacyJit |      X86 |  False |  1,030.2 ms |     58.91 ms |     9.132 ms |
  DedicatedParallelIo4 | LegacyJit |      X86 |  False |  1,005.3 ms |     69.99 ms |    10.850 ms |
                 SeqIo | LegacyJit |      X86 |   True | 45,701.7 ms | 17,144.07 ms | 2,657.821 ms |
 DedicatedParallelIo16 | LegacyJit |      X86 |   True |  1,307.2 ms |  1,375.45 ms |   213.235 ms |
  DedicatedParallelIo8 | LegacyJit |      X86 |   True |  1,093.7 ms |    956.08 ms |   148.220 ms |
  DedicatedParallelIo4 | LegacyJit |      X86 |   True |    936.3 ms |     60.61 ms |     9.396 ms |
                 SeqIo |    RyuJit |      X64 |  False | 47,281.4 ms | 28,170.13 ms | 4,367.176 ms |
 DedicatedParallelIo16 |    RyuJit |      X64 |  False |  1,183.5 ms |    614.91 ms |    95.328 ms |
  DedicatedParallelIo8 |    RyuJit |      X64 |  False |  1,204.7 ms |    519.85 ms |    80.592 ms |
  DedicatedParallelIo4 |    RyuJit |      X64 |  False |  1,220.7 ms |    392.94 ms |    60.917 ms |
                 SeqIo |    RyuJit |      X64 |   True | 46,977.9 ms | 12,592.21 ms | 1,952.153 ms |
 DedicatedParallelIo16 |    RyuJit |      X64 |   True |  1,084.6 ms |    185.04 ms |    28.687 ms |
  DedicatedParallelIo8 |    RyuJit |      X64 |   True |  1,158.7 ms |    990.17 ms |   153.504 ms |
  DedicatedParallelIo4 |    RyuJit |      X64 |   True |  1,089.3 ms |    197.70 ms |    30.648 ms |
                 SeqIo |    RyuJit |      X86 |  False | 48,314.5 ms | 35,613.41 ms | 5,521.097 ms |
 DedicatedParallelIo16 |    RyuJit |      X86 |  False |  1,217.6 ms |    172.36 ms |    26.720 ms |
  DedicatedParallelIo8 |    RyuJit |      X86 |  False |  1,246.7 ms |    280.16 ms |    43.434 ms |
  DedicatedParallelIo4 |    RyuJit |      X86 |  False |  1,217.2 ms |    338.57 ms |    52.489 ms |
                 SeqIo |    RyuJit |      X86 |   True | 45,880.3 ms | 17,490.26 ms | 2,711.491 ms |
 DedicatedParallelIo16 |    RyuJit |      X86 |   True |  1,018.4 ms |    498.71 ms |    77.314 ms |
  DedicatedParallelIo8 |    RyuJit |      X86 |   True |  1,042.4 ms |    607.26 ms |    94.143 ms |
  DedicatedParallelIo4 |    RyuJit |      X86 |   True |  1,011.4 ms |    681.67 ms |   105.679 ms |
```

</p></details>

Saying that this PR closes #179 and #167 as well

Feel free to ask from more tests/refactoring whatever. This took lot of time so I'm afraid I might missed some debugging artifacts in code - I mean this requires thorough review. Also I'm not 100% satisfied with code quality, I did lot of logic duplication in some places - we can elaborate on code reorganization if necessary.

[All benchmarks I used](https://github.com/Rattenkrieg/fscheckbenchmarks)

src/FsCheck/Testable.fs
```diff
@@ -32,35 +24,105 @@ type Result =
...
+    ///Returns a new result that is Succeeded if and only if both this
+    ///and the given Result are Succeeded.
+
+    static member (&&&) (l,r) = 
+        //printfn "And of l %A and r %A" l.Outcome r.Outcome
+        match (l.Outcome,r.Outcome) with
+        | (OutcomeContainer.Future tl,OutcomeContainer.Future tr) -> 
+            { 
+            Outcome = OutcomeContainer.Future <| Threading.Tasks.Task.WhenAll([|tl; tr|]).ContinueWith(fun (xs :Threading.Tasks.Task<Outcome[]>) -> Result.combineArray (&&&) xs.Result)
+            Stamp = [] // that superlame I'm afraid I had to place task at `Value of Result / Future of Task<Result>` level
```
Rattenkrieg: nuff said
I'm really don't want to push task on higher level, blocking on `Result` here seems viable option

kurtschelfthout:
Nice, thanks a lot. I will take a look! It may take some time, have a lot going on at the moment.

kurtschelfthout:
Sorry for the delay, have not forgotten about this! I will take a look shortly (i.e. this week).

Rattenkrieg:
No problem, feel free to ask about additional comments (whether in code or here) if you feel some places are not clear enough.

docs/csharp/TestData.cs
```diff
@@ -1,9 +1,5 @@
 ﻿using FsCheck;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
```
kurtschelfthout: There are a bunch of cleanup related changes, mostly in docs. Worth just submitting these as a separate PR to master?

src/FsCheck/Runner.fs
```diff
@@ -33,6 +33,10 @@ type IRunner =
     ///Called whenever all tests are done, either True, False or Exhausted.
     abstract member OnFinished: string * TestResult -> unit
 
+[<NoEquality;NoComparison>]
```
Rattenkrieg: Dunno, just copy-pasted from somwhere in this file. ¯\_(ツ)_/¯
Will remove.

src/FsCheck/Runner.fs
Rattenkrieg: > I guess the option type also encodes whether or not to do a parallel run

Right.
Regarding `MaxDegreeOfParallelism` I see only two reasonable options: `Environment.ProcessorCount` - for cpu bound crunching and 1 - for async i/o bound work (however `Environment.ProcessorCount` is also ok for i/o). Overall this parameter may confuse users, but I don't want to left them with hardcoded defaults. So I'm going to put dense comments in code and suggest to give a couple of lines regarding this in documentation (I can contribute to it by my own).
If you have different vision on configuration style - I'm ok to stick with yours.

src/FsCheck/Random.fs
```diff
@@ -151,44 +152,47 @@ type internal Random () =
...
                 while not result.HasValue do
                     if candidate < l || candidate >= h then 
-                        candidate <- Random.nextInt (s, &s')
+                        candidate <- Random.nextInt (&s'')
                     else
                         result <- System.Nullable<_> (candidate)
 
             // Return the result.
+            s' <- s''
             result.Value
```
Rattenkrieg: Actually there was one #357 
I will rebase on it.

tests/FsCheck.Test/Runner.fs
```diff
@@ -23,6 +116,7 @@ module Runner =
         underTest <= 0.0
 
     [<Fact>]
+    // TODO: Arguments exhausted after 0 tests.
```
Rattenkrieg: This works now, just rudiments, sorry. 🙈

tests/FsCheck.Test/Runner.fs
```diff
@@ -202,7 +296,7 @@ module BugReproIssue344 =
             
         // run this on another thread, and abort it manually
         let thread2 = Thread(fun () ->
-            let config = { Config.Quick with Arbitrary = [ typeof<MyGenerators> ]  }
+            let config = { Config.Quick with Arbitrary = [ typeof<MyGenerators> ]; ParallelRunConfig = None  }
```
Rattenkrieg: As far as I recall there were several issues (forgot which exactly) with running this scenario in parallel mode. For now It runs ok in parallel. Also note that I left parallel mode as [default](https://github.com/fscheck/FsCheck/pull/351/files/fac6f6b7a1cccb6420bcf3768cc47d66ed24100f#diff-fae3712c738f2860b54d6ce6f880b74aR536)  for now. Just to see how it will behave on CI server, will remove before merge.

docs/csharp/TestData.cs
Rattenkrieg: I've accidently hit lightbulb in VS and ended with global cleanup. :rage1: 
If you dont't like to see this changes in this particular commit I can revert it manually.

src/FsCheck/Testable.fs
Rattenkrieg: I left it empty intentionally with hope for your assistance.
With `Outcome` being wrapped in `Task` I can combine these fields only by accessing `Task<Outcome>.Result`, hence blocking. If these operators are not used intensively in generation/testing (what I can't tell for sure) maybe it will be ok to simply put calls to `Task<Outcome>.Result`.

src/FsCheck/Runner.fs
kurtschelfthout: The config setup is fine with me. Yes, happy to take docs on configuration etc.

src/FsCheck/Testable.fs
kurtschelfthout: `Stamp` is used to gather strings given to `collect` and friends that are used to observe test case distribution; `Label` for strings that are used with the `label` function to show output on failure; and `Arguments` is used to collect counter-examples. So they are used often. But I guess the question is whether the `&&&` and `|||` combinators are used often - I think they are relatively uncommon. 

Perhaps the `OutcomeContainer` should be a `ResultContainer` so the whole thing can be wrapped in a future; or we can use blocking on `Result` with the only downside (?) being that users of `&&& ` and `|||` will be potentially slowed down.

src/FsCheck/Random.fs
kurtschelfthout: Ah yes, of course. Yes, please rebase.

docs/csharp/TestData.cs
kurtschelfthout: No, it's fine. Bigger fish to fry.

src/FsCheck/Testable.fs
Rattenkrieg: > Perhaps the OutcomeContainer should be a ResultContainer so the whole thing can be wrapped in a future; or we can use blocking on Result with the only downside (?) being that users of &&& and ||| will be potentially slowed down.

That exactly the options which I'm considering for now. I'm afraid wrapping result in a future may end up with bigger mess and code duplication due to "async all the way down", and maybe we would found ourselves in front of another call requiring blocking access somwhere up the call chain. ☹️  
To block at `&&&` and `|||` seems seductive for me but somewhat unfair regarding users, so I'll try to hack quick solution with first approach and left blocking as fallback option.

Rattenkrieg:
After rebasing on top of #383 performance dropped twice ._.
Two runs before and after rebase:

default

Rattenkrieg:
To be more precise performance is affected when changing [this](https://github.com/Rattenkrieg/FsCheck/blob/c8fbaf13c12f0b5082952bad92ff4075cdb06dbd/src/FsCheck/Runner.fs#L161) [lines](https://github.com/Rattenkrieg/FsCheck/blob/c8fbaf13c12f0b5082952bad92ff4075cdb06dbd/src/FsCheck/Runner.fs#L207) to
`let (MkRose (Lazy result, shrinks)) = Gen.eval (size |> round |> int) rnd gen`

kurtschelfthout:
Yes - This is expected. That change has been on master for a while and causes the size of generated structures to increase significantly. In other words, FsCheck has to work harder than before, but has less chance of generating many duplicate, small values (e.g. Empty lists). Hopefully this does not affect the relative speed gain comparing sequential and parallel generation?

Rattenkrieg:
> Hopefully this does not affect the relative speed gain comparing sequential and parallel generation?

I will run benchmarks before pushing rebased commit.
Although it's ready for review - All fixes regarding your comments are done. Task is lifted at `ResultContainer` level.

src/FsCheck/Runner.fs
```diff
@@ -204,7 +204,7 @@ module Runner =
     let private testStep rnd (size :float) gen =
             let result, shrinks =
                 try
-                    let (MkRose (Lazy result, shrinks)) = Gen.eval (size |> round |> int) rnd gen
+                    let (MkRose (Lazy result, shrinks)) = (eval (newSize |> round |> int) rnd2).Value
```
Rattenkrieg: I believe `newSize` and `rnd2` have to be `size` and `rnd`

kurtschelfthout:
Yes, there were a few other build problems too which I've fixed, but I can't push? There should be a switch on the PR to allow me to push to your branch, is that set?

> git push --set-upstream parallel Rattenkrieg-parallel-run-83
Warning: Permanently added 'github.com,192.30.253.112' (RSA) to the list of known hosts.
Counting objects: 7, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (7/7), done.
Writing objects: 100% (7/7), 669 bytes | 0 bytes/s, done.
Total 7 (delta 6), reused 0 (delta 0)
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
To github.com:Rattenkrieg/FsCheck.git
 ! [remote rejected] Rattenkrieg-parallel-run-83 -> Rattenkrieg-parallel-run-83 (permission denied)
error: failed to push some refs to 'git@github.com:Rattenkrieg/FsCheck.git'

kurtschelfthout:
Ok, we're getting somewhere. At least the normal build now works on AppVeyor. The .NET Standard build however fails with:
```
C:\projects\fscheck\src\FsCheck\Runner.fs(293,34): error FS0039: The field, constructor or member 'Yield' is not defined. [C:\projects\fscheck\src\FsCheck.netcore\FsCheck.netcore.fsproj]
```

Somehow it doesn't know about `Thread.Yield()`? Maybe they removed that in .NET Standard or something?

tests/FsCheck.Test/Runner.fs
```diff
@@ -1,5 +1,150 @@
...
+        let runner2 = ProbeRunner () 
+        FsCheck.Runner.check { config with Replay = seed; Runner = runner2; ParallelRunConfig = None } p
+        runner1.Equals runner2
+            
+    [<Fact>]
+    let ``parallelTest produces same sequence as test on success`` () =
+        let body i = true
+        let arb = Arb.from<int>
+        let config = { Config.Quick with MaxTest = 30000; EndSize = 30000 }
+        if not <| check arb body config then failwith "can't find the way to call xunit assertion lol"
```
kurtschelfthout: Most of the tests use Unquote, `open Swensen.Unquote` then you can use `test <@ boolean assertion @>`. But I'm actually tiring of all these dependencies for very little benefit, esp. since Unquote doesn't have .NET standard build yet for example, so I'm happy with just throwing an exception like you did.

tests/FsCheck.Test/Runner.fs
```diff
@@ -1,5 +1,150 @@
...
+
+    [<Property(Replay = "10538531436017130025,14826463994991344553", MaxTest = 30000, EndSize = 30000)>]
+    let ``30k ints fails`` () =
+        true
+        
+    [<Fact>]
+    let ``bad case`` () = 
+        let rnd = Random.createWithSeedAndGamma (13471455474525100574UL,7555858534656909083UL)
+        let n = 19938
+        Random.rangeInt (0, n, rnd)
```
kurtschelfthout: What are these two tests checking for?

kurtschelfthout:
> Although it's ready for review - All fixes regarding your comments are done. Task is lifted at ResultContainer level.

I think it's ready for merge, once the build works. I'm sure there will be some issues but we can take them as they come up. I think the approach overall is sound - split the seed nb of tests times, then run each test in a separate task. I'm surprised TPL is not better at this - I would think this is what a parallel for is ideal for? Any thoughts on why that didn't work?

tests/FsCheck.Test/Runner.fs
Rattenkrieg: At some moment my implementation was failing and these tests are repro for such case. We don't need it anymore, I will remove it.

Rattenkrieg:
I ran benchmarks comparing sequential testing in 7e6122d0711345d0796a240c633787e9e110c645 and my branch rebased on top of it to make sure I haven't introduced any degradation to sequential scenario. Timings seems identical. Relative speedup of parallel version haven't changed as well.

>  I'm surprised TPL is not better at this - I would think this is what a parallel for is ideal for? Any thoughts on why that didn't work?

 - TPL parallel routines and PLINQ are super greedy regarding memory
 - It doesn't provide much ways to control work distribution
 - I haven't found convenient way to bail out quickly when test was falsified

tests/FsCheck.Test/Runner.fs
Rattenkrieg: Oh I recalled these tests are for #356 
They already shipped in that PR so removing it here as duplicates.

paket.dependencies
```diff
@@ -11,4 +11,6 @@ nuget Unquote
 nuget xunit
 nuget xunit.runner.console
 nuget xunit.runner.visualstudio version_in_path: true
-nuget FSharp.Formatting
\ No newline at end of file
+nuget FSharp.Formatting
+nuget Expecto
+nuget System.Threading
```
Rattenkrieg: > C:\projects\fscheck\src\FsCheck\Runner.fs(293,34): error FS0039: The field, constructor or member 'Yield' is not defined. [C:\projects\fscheck\src\FsCheck.netcore\FsCheck.netcore.fsproj]

Simply adding System.Threading don't work. Have no idea what to do.

examples/FsCheck.NUnit.Examples/PropertyExamples.fs
```diff
@@ -10,10 +10,18 @@ type NUnitTest() =
     [<Test>]
     member __.NormalTest() =
         ignore true
-
+        
     [<Property>]
     member __.RevUnit (x:char) =
         List.rev [x] = [x]
+
+    // Parallelism for PropertyAttribute not implemented for Nunit yet
```
Rattenkrieg: I did it for Xunit using `Replay` as example. But for Nunit `Replay` isn't implemented, and I have no idea for what reason? Are there any restrictions on Nunit side? Should I implement parallelism and replay there?

paket.dependencies
kurtschelfthout: Ah, I think I know what's going on there: the netcore project files are not managed via paket (yet, I may change this in the future, but for now paket doesn't yet do exactly what I want with .NET standard). So instead of changing paket.dependencies try adding that package to FsCheck.netcore.fsproj.

examples/FsCheck.NUnit.Examples/PropertyExamples.fs
kurtschelfthout: Probably just because no one did it yet. But let's keep those changes for later PRs; this one is big enough as it is. If you do decide to add Replay for NUnit then please do that on master first; then I'll merge it back here, and after that merge hits the fscheck3 branch we can add parallellism to it.

Rattenkrieg: Got it, I will send PR to master.

src/FsCheck/Runner.fs
```diff
@@ -79,50 +86,267 @@ module Runner =
...
+            else
+                subE.Dispose (); false    
+        let moveNextOuter () =
+            if i = maxTest && firstRun then
+                results <- Array.zeroCreate<seq<TestStep>> maxFail
+                i <- 0
+                run unluckySeq indexF
+                firstRun <- false
+            while isNull results.[i] do
+#if NETSTANDARD1_6
```
Rattenkrieg: I just figured out that there is no `Yield()` for netstandard1.6

kurtschelfthout:
Merged! Awesome.

PR #354, opened 2017-03-01
fix #324

### LINKED ISSUE #324: VS-Integration: FsCheck.XUnit does not show currently executing test
When I run a XUnit Property test, the test is greyed out and when it finishes, it gets either the green check or the red icon:

image

image

When I run a long running Fact, I can see an in-progress icon:

image

image

Since Property tests can potentially take a long time, it would be great to see which test is currently executing.

0x53A:
ref https://github.com/fscheck/FsCheck/issues/324

### DIFF
--- a/src/FsCheck.Xunit/PropertyAttribute.fs
+++ b/src/FsCheck.Xunit/PropertyAttribute.fs
@@ -231,6 +231,9 @@ type PropertyTestCase(diagnosticMessageSink:IMessageSink, defaultMethodDisplay:T
                 cancellationTokenSource.Cancel() |> ignore
             summary
 
+        if not (messageBus.QueueMessage(new TestCaseStarting(this))) then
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Thanks for taking this on! Let us know when this is ready...

0x53A:
I think it is ready, I just havent gotten around to testing it yet.

0x53A:
It looks good:
image

After a test finishes the icon also correctly switches to green / red.

--------------------------

I just took one more look at the xunit source, and there is also a TestCaseFinished (https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.execution/Sdk/Frameworks/Runners/TestCaseRunner.cs#L93)

For completeness' sake I also added that, but I have no idea what (if any) effect this message has.
In a quick test it at least didn't seem to break anything...

Issue #356, opened 2017-03-04
Random.rangeInt hangs


Rattenkrieg:
Stumbled upon when running benchmarks for #351 

Two ways to reproduce:
```fsharp
    [<Property(Replay = "10538531436017130025,14826463994991344553", MaxTest = 30000, EndSize = 30000)>]
    let ``30k ints fails`` () =
        true

    [<Fact>]
    let ``bad case`` () = 
        let rnd = Random.createWithSeedAndGamma (13471455474525100574UL,7555858534656909083UL)
        let n = 19938
        Random.rangeInt (0, n, rnd)
```

So it appears that [this loop](https://github.com/fscheck/FsCheck/blob/fscheck3/src/FsCheck/Random.fs#L180) has fixed points (if you treat it like function)
OR
Maybe `Random.nextInt (s, &s')` should be called with arguments swapped each time?

kurtschelfthout:
> Maybe Random.nextInt (s, &s') should be called with arguments swapped each time?

You're totally right, that isn't right. The `s` should be updated to `s'` after each iteration.

Here was the difference with the original nicely functional (but slow) code:
https://github.com/fscheck/FsCheck/commit/3317199e63d35a201e9c364e7c1eed2703ff4cdd

Rattenkrieg:
Ok, it seems I figured out correct pattern. Will provide fix now.

PR #357, opened 2017-03-04
Random now correctly updates rng (fix for #356)

Rattenkrieg:
Found no usage of first param to `next*` functions, so I have removed it.
Just to note - F# compiler surprisingly not giving error when I'm not assigning value to out param.


kurtschelfthout:
Awesome, thanks!

Yeah, the F# compiler is not so polished with byref, out, default values, unsafe, p/invoke and such things.

PR #360, opened 2017-03-09
Updated xUnit.net to version 2.2.0

ploeh:
The update was done using `paket update`, so I can't explain all the
changes done to various files...


ploeh:
Clearly not a success 👎

ploeh:
It's getting too late... No more yak shaving today for my part.

kurtschelfthout:
Puzzling. And what is .NET Standard 1.6.**1** ? As if things aren't confusing enough, MSFT seems to be constantly messing with the versions...

Issue #367, opened 2017-04-24
support C# DTO in reflection based generators


giacomociti:
we currently support (automatically deriving generators with refection) c# record types of this shape:

```
class MyRecord
    {
        public readonly int I;
        public readonly string S;
        public MyRecord(int i, string s)
        {
            I = i;
            S = s;
        }
    }

```
unfortunately, most libraries and frameworks encourage DTO with the following shape:

```

class MyDto
    {
        public int I { get; set; }
        public string S { get; set; }
    }

```
It would be nice to enhance FsCheck to include this kind of data object.
Since FsCheck now uses LINQ expression trees, this should be easy to achieve, I may be able to contribute this.

mausch:
Hi, the problem with this is it adds some ambiguity to the generation logic, as sometimes mutable properties are treated as optional values. 
For example, given this class:

```c#
class Foo {
  public int? Number { get; set; }
  public string Name { get; }
  public Foo(string name) { this.Name = name; }
}
```

Should FsCheck set `Number` or not? There may be reasons both for ("it is possible to set it so it should be covered") and against setting it ("it was meant to be optional") so generation becomes harder to reason about. 

I'd keep reflection-based generation simple in FsCheck, and try to create a separate reflection-based generator for such classes, perhaps using [TypeShape](https://github.com/eiriktsarpalis/TypeShape) (I don't know if it handles mutable properties), which you would then explicitly register in FsCheck.

kurtschelfthout:
I like this suggestion; it's a request that returns often. 

If we only allow objects with a default constructor, and set all the setters with a random value implied by their type, I don't think that would create a lot of ambiguity? And this is a pattern that is used for attributes for example.

I agree with @mausch that "mixed" approaches that both have a constructor and setters are harder to support well, but I don't feel that should stop us from supporting the 80% case out of the box.

giacomociti:
I just discovered this [snippet](http://www.fssnip.net/7RA/title/Generic-POCO-generators-for-FsCheck) showing TypeShape in action.

kurtschelfthout:
Released in https://github.com/fscheck/FsCheck/releases/tag/2.8.1

PR #369, opened 2017-04-25
support C# DTO in reflection based generators

giacomociti:
this PR is related to #367: support for DTO classes is kept separate from existing support for C# record classes.
A DTO is a class with the default (parameterless) constructor as the only public constructor and having some writable public property. For each writable public property a randomly generated value is set, using a member init expression.
This PR may be enhanced with some unit test (I could not find any existing test related to reflection based generation) and also adjusting docs (where it is stated that only immutable classes are supported).


kurtschelfthout:
Thanks! This is looking good.

> This PR may be enhanced with some unit test (I could not find any existing test related to reflection based generation)

Yes, please. There are a few tests for "C# record type" generation, start here: https://github.com/fscheck/FsCheck/blob/master/tests/FsCheck.Test/Arbitrary.fs#L520

> and also adjusting docs (where it is stated that only immutable classes are supported).

Yes, please if you know where that is, I couldn't actually immediately find it so don't blame you :)

PR #375, opened 2017-08-13
Add negativeint

stijnmoreels:
Several times in an 'unhappy' scenario, I needed negative integers. So it would be nice to have them default within FsCheck!

Not sure though if I changed "enough" to made the generator discover the new type?

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -12,6 +12,7 @@ open System.Net
 open System.Net.Mail
 #endif
 
+///Represents and int < 0
 type NegativeInt = NegativeInt of int with
     member x.Get = match x with NegativeInt r -> r
     override x.ToString() = x.Get.ToString()
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -801,7 +801,7 @@ module Arb =
 
         static member NegativeInt() =
             Default.Int32()
-            |> mapFilter ((*) -1) (fun x -> x < 0)
+            |> mapFilter (fun x -> -abs x) (fun x -> x < 0)
             |> convert NegativeInt int
 
         static member NonNegativeInt() =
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
Strange that the build fails. Looks like an infinite loop/stackoverflow somewhere.

stijnmoreels:
Is it possible something I missed?

kurtschelfthout:
I don't know, in that case I'm missing it too! :) But it does seem to have something to do with this change. I re-ran the tests on master and they work. If you run tests locally everything works?

stijnmoreels:
:) I think I found it. Stupid mistake, while reusing the 'PositiveInt' type, I copied the 'abs' also ;).

kurtschelfthout:
Ah yes, good call, totally missed that as well! I just have one comment and then this is good to go.

src/FsCheck/Arbitrary.fs
```diff
@@ -793,6 +799,11 @@ module Arb =
                             yield new KeyValuePair<_,_>(key, value) }
             fromGenShrink(genKeyValuePair,shrinkKeyValuePair)
 
+        static member NegativeInt() =
+            Default.Int32()
+            |> mapFilter ((*) -1) (fun x -> x < 0)
```
kurtschelfthout: This should be `|> mapFilter (fun x -> -abs x) (fun x -> x < 0)` ideally. 

The map (first function) is really an optimisation - the filter (second function) is applied regardless so even if the map doesn't work you won't see a problem in the generated values. But a good map decreases the number of values that are being generated then thrown away by the filter.

Your map function just flips the sign, so it isn't very effective in making a positive or negative int into a negative int. My suggestion above makes them all negative except 0s (which will still be filtered out by the filter)

stijnmoreels: Ah, yes; makes sense, thanks!

kurtschelfthout:
Thanks!

PR #378, opened 2017-08-23
use SourceLink version 2

### LINKED ISSUE #377: Use SourceLink (again)
I removed sourcelink in https://github.com/fscheck/FsCheck/commit/68577d2c7c11406f3e93bc996e1bb3e6ed55f4ed and https://github.com/fscheck/FsCheck/commit/f5ae9055defce9bf687ac9aacd3dd9e1e7b10edc because it kept throwing errors about not finding some Dataflow library. I don't know exactly why (dotnet update? VS2017?) Tried getting SourceLink v2 to work but couldn't.

Also dislike the way SourceLink v2 needs to be set up, would prefer to manage everything via paket and `build.fsx` instead of relying on an msbuild 15 feature.

Also happy to use an alternative if it's well supported. Perhaps symbol server like in the old days?

I won't be working on this myself, because frankly I'm exasperated with keeping up with all the changes already.

ctaggart:
fix #377

This follows the [quick start](https://github.com/ctaggart/SourceLink#quick-start) by:
* adding a Directory.Builds.props
* building with MSBuild 15
* calling `mbuild /t:restore` so that the SourceLink MSBuild targets are hooked in
* switching to Portable PDB files

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+image: Visual Studio 2017
 init:
   - git config --global core.autocrlf input
 build_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,8 @@ image: Visual Studio 2017
 init:
   - git config --global core.autocrlf input
 build_script:
+  - msbuild restore FsCheck.sln
+  - msbuild restore FsCheck.netcore.sln
   - cmd: ./build.cmd CI
 test: off
 version: '{build}'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,8 @@ image: Visual Studio 2017
 init:
   - git config --global core.autocrlf input
 build_script:
-  - msbuild restore FsCheck.sln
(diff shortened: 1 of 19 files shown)


kurtschelfthout:
Ooooh. You should really note somewhere that DebugType needs to be portable :) ...or I'm an idiot for totally missing that.

ctaggart:
I'll try to make it even more clear in the docs by updating the example Directory.Build.props with what I used here.

kurtschelfthout:
I've had to revert this because it was making debugging impossible. With DebugType=portable VS 2017 doesn't seem to have any idea where in the source code things actually are. E.g when an exception is thrown, the text that is highlighted in the debugger is a completely different part of the code.

Sorry to be such a harsh critic as I know you put a lot of effort into this. But with all the other ecosystem changes that are going on I don't want to spend the time now. So for the moment I'm just going to try to publish source and symbol packages instead, by enabling that in `packet pack`, and if that doesn't work, then I'll settle for nothing.

Issue #387, opened 2017-09-09
Test "parallelTest works with custom arb" fails with Stackoverflow


kurtschelfthout:
Possibly has something to do with new size mechanism and Gen.piles introduced after merge from master to fscheck3.

Rattenkrieg:
Which one fails? I got everything green after rebase.

kurtschelfthout:
Think it only happens with release build. Run the build script to see.

Rattenkrieg:
I've found failing one - `parallelTest works with custom arb`
I tend to think SO is related to custom arbitrary rather than parallelism. I've expierenced similar issue (in sequntial tests) with custom arbitrary for predefined types like ints, strings. Will investigate soon.

kurtschelfthout:
Yeah, got a stack trace at some point. Lots of Gen.map closures. Could be it's reproducible just by generating a long list or something. 

I only get it when running with the console runner though, which is weird.

PR #388, opened 2017-09-10
Replay single test

Rattenkrieg:
Fix for #310 
Will edit docs in this PR after we agree about api.

### DIFF
--- a/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
+++ b/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
@@ -15,13 +15,12 @@ type NUnitTest() =
     member __.RevUnit (x:char) =
         List.rev [x] = [x]
 
-    // Parallelism for PropertyAttribute not implemented for Nunit yet
-    // [<Property(Parallelism = 8)>]
-//    member __.Async (i:int) =
-//        async { 
-//            do! Async.Sleep 1500
-//            return true
-//        }
+    [<Property(Parallelism = 8)>]
+    member __.Async (i:int) =
+        async { 
+            do! Async.Sleep 1500
+            return true
+        }
   
     [<Property>]
     member __.RevApp (x:string) xs =
@@ -58,3 +57,7 @@ type NUnitTest() =
     [<Property(QuietOnSuccess = true)>]
     member __.NoOutputOnSuccess (x:char) =
         List.rev [x] = [x]
+
+    [<Property( Replay="7797334217889672160,158684412801592049,91", Verbose = true )>]
+    member __.Get82_shouldFail (i:int) =
+        i < 80
--- a/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
+++ b/examples/FsCheck.NUnit.Examples/PropertyExamples.fs
@@ -57,7 +57,3 @@ type NUnitTest() =
     [<Property(QuietOnSuccess = true)>]
     member __.NoOutputOnSuccess (x:char) =
         List.rev [x] = [x]
-
-    [<Property( Replay="7797334217889672160,158684412801592049,91", Verbose = true )>]
-    member __.Get82_shouldFail (i:int) =
-        i < 80
--- a/src/FsCheck/Runner.fs
+++ b/src/FsCheck/Runner.fs
@@ -18,7 +18,9 @@ type TestResult =
                 * list<obj>(*the original arguments that produced the failed test*)
                 * list<obj>(*the shrunk arguments that produce a failed test*)
                 * Outcome(*possibly exception or timeout that falsified the property*) 
-                * Rnd (*the seed used*)
+                * Rnd (*initial seed for test run*)
+                * Rnd (*the seed used for generation of arguments that falsified test*)
+                * int (*the size used for generation of arguments that falsified test*)
     | Exhausted of TestData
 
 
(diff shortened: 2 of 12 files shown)


examples/FsCheck.NUnit.Examples/PropertyExamples.fs
```diff
@@ -57,4 +56,8 @@ type NUnitTest() =
 
     [<Property(QuietOnSuccess = true)>]
     member __.NoOutputOnSuccess (x:char) =
-        List.rev [x] = [x]
\ No newline at end of file
+        List.rev [x] = [x]
+
+    [<Property( Replay="7797334217889672160,158684412801592049,91", Verbose = true )>]
```
Rattenkrieg: Thats how api looks like. 91 is for test case to fast-forward.

kurtschelfthout:
I think instead `Replay` should contain the Rnd for the failing test specifically and the `int` size for that test. We should record this in TestStep right before running that step and report it in the output if the test fails. And then replay just evals that one test directly with the given Rnd and size.

Rattenkrieg:
@kurtschelfthout isn't [that](https://github.com/fscheck/FsCheck/pull/388/files#diff-fae3712c738f2860b54d6ce6f880b74aR36) exactly what you stated? And [this is](https://github.com/fscheck/FsCheck/pull/388/files#diff-fae3712c738f2860b54d6ce6f880b74aR193) how we run it. Sorry if I haven't got your point properly.

kurtschelfthout:
No, don't think so, but I'm about to board a plane so have only taken a quick look so far. But see #389 for where I got with this a while back.

Rattenkrieg:
It's completely clear for me now. Will share my thoughs here soon.

Rattenkrieg:
From my subjective point of view it would be potential source of confusion for users to pass new seed and gamma pair along with test step. Just appending some number after comma seems more straightforward. But thats completely IMHO. If you see interface not confusing I'm ok to rework PR with your approach.

kurtschelfthout:
It's the exact same interface - just the meaning of the third number changes from nth test to size of the test at that time.

But that makes it more robust to changes in FsCheck test runner (which may use seeds in a different way or change the size over the test runs in a different way), and allows a true fast forward (because Gen is really just a function from Rnd and size to value), you don't even need to split.

Rattenkrieg:
Ok, will fix #387 first and then rework this according to your vision.

src/FsCheck/Runner.fs
```diff
@@ -150,45 +158,45 @@ module Runner =
...
-                yield! test' newSize resize rnd1 gen
-            }
+    let stepsSeq resize =
+        Seq.unfold (fun (initSize, rnd) ->
+            let rnd1, rnd2 = Random.split rnd
+            let newSize = resize initSize
+            Some ((rnd1, rnd2, newSize), (newSize, rnd1)))
+
+    let private test initSize resize rnd0 gen =    
+        let test' ((Gen eval) as gen) (rnd1, rnd2, newSize) =
```
Rattenkrieg: I was pursuing goal to keep this function applicable to regular test run as well as replay. That led to ugly "two-cell memory" with `preLastSeed` and `preLastSize` in `runner`. Looking at what being done I don't know whether that was good idea. We need those `preLastXXX` because initial size and seed are not used to produce args, only for seeding in `stepsSeq`, so we have to report rnd and size previous to those led to fault which will be seeded in `stepsSeq`. So maybe exposing inner `test'` for replaing purposes is viable alternative.

src/FsCheck/Runner.fs
```diff
@@ -377,19 +380,27 @@ module Runner =
         let tryShrinkNb = ref 0
         let origArgs = ref []
         let lastStep = ref (Failed Res.rejectedV)
-        let seed = match config.Replay with None -> Random.create() | Some s -> s
+        let seed, size = match config.Replay with None -> Random.create(), None | Some s -> s.Rnd, s.Size
         let increaseSizeStep = float (config.EndSize - config.StartSize) / float config.MaxTest
+        let preLastSeed = ref seed
+        let preLastSize = ref (int ((float <| defaultArg size config.StartSize) - increaseSizeStep))
+        let lastSeed = ref seed
+        let lastSize = ref (int (float <| defaultArg size config.StartSize))
```
Rattenkrieg: Well, coding at 3 am weekdays never a good idea. If I'd choose to separate `test'` that ugliness wouldn't be here. Will hack another approach today.

src/FsCheck/Runner.fs
```diff
@@ -377,19 +380,27 @@ module Runner =
         let tryShrinkNb = ref 0
         let origArgs = ref []
         let lastStep = ref (Failed Res.rejectedV)
-        let seed = match config.Replay with None -> Random.create() | Some s -> s
+        let seed, size = match config.Replay with None -> Random.create(), None | Some s -> s.Rnd, s.Size
```
Rattenkrieg: It looks stupid to keep replay size in separate field when we already have `config.StartSize` but nevertheless we have to indicate that this is "replay run" somehow.

Rattenkrieg:
@kurtschelfthout I think it's ready for review.

kurtschelfthout:
Awesome, looks broadly good, will take a more detailed look soon.

PR #405, opened 2017-10-12
Ignored Paket.Restore.targets

ploeh:
According to
https://github.com/fsprojects/Paket/issues/2840#issuecomment-336080872
it should be ignored by Git:

> with latest paket you can gitignore it

- @forki

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,5 @@ docs/tools/XmlWriter
 project.lock.json
 .vscode
 
+# Paket
+.paket/Paket.Restore.targets


kurtschelfthout:
Does the latest paket somehow download that file? It used to be only applicable to .NET core/standard projects, for which we're not using paket anyway.

forki:
It restores that file yes. And why don't you use paket for netcore?

kurtschelfthout:
> why don't you use paket for netcore?

For the most part just that no one got to it, original PR was using dotnet CLI only as paket didn't work for netcore at the time I believe.

I tried changing at one point, and disliked that paket added conditional references to all project files to include PCL/netstandard/netframework, while we have specific project files for each. This complicates merges esp since we now have master for ongoing v2 dev and fscheck3 for v3 dev. 

I started playing around with framework restrictions and groups but ended up thinking perhaps we should just release netstandard only for v3 to simplify the whole thing.

Sorry for braindumping on you :laughing:

forki:
> and disliked that paket added conditional references to all project files to include PCL/netstandard/netframework

for new sdk projects? we don't add references at all any more.

Look at https://github.com/SAFE-Stack/SAFE-BookStore/blob/master/src/Server/Server.fsproj#L22 that's the only paket line that you will see in a fsproj

kurtschelfthout:
> for new sdk projects? we don't add references at all any more.

Yes, and that is awesome, but I meant in the existing traditional projects.

forki:
Ok I'm confused. You want to do netcore/netstandard with old school fsproj?

Am 12.10.2017 17:29 schrieb "Kurt Schelfthout" <notifications@github.com>:

kurtschelfthout:
> Ok I'm confused. You want to do netcore/netstandard with old school fsproj?

Sorry. Let me try to explain.

We use paket currently for everything - pulling down packages and packing - for the old school builds for netframework and PCL , and which have old school project files.

We don't use paket for netcore - mostly because that's just the way it was initially set up - but for those we of course use the new style project files.

Unless the situation has changed recently, I don't want to switch everything to new project files because that makes development in VS impossible.

The paket behavior I observed when trying to also use paket for the netstandard projects, is that it also added netstandard references to the old school project files that target PCL and/or netframework.

So we have:
- FsCheck.fsproj: uses paket, old school, builds netframework target
- FsCheck.Profile7.fsproj: uses paket, old school, builds PCL target
- FsCheck.netcore.fsproj: does not use paket, new style project, builds netstandard target

And when I tried to add paket references to FsCheck.netcore.fsproj, that resulted in netstandard/netcore references to be added to FsCheck.fsproj and FsCheck.Profile7.fsproj, or something along those lines.

forki:
Vs 15.5 preview came out today. Things will improve eventually. But I hear
you

Am 12.10.2017 18:07 schrieb "Kurt Schelfthout" <notifications@github.com>:

Issue #410, opened 2017-11-03
Installing NuGet package on net461 framework installs superfluous dependencies


hickford:
Installing FsCheck with nuget.exe for .NET Framework 4.6.1 installs a bunch of extra packages that aren't required, in particular NETStandard.Library:

    > nuget install FsCheck -Framework net461
    ...
    Retrieving package 'FsCheck 2.10.3' from 'nuget.org'.
    Retrieving package 'FSharp.Core 4.1.18' from 'nuget.org'.
    Retrieving package 'Microsoft.NETCore.Platforms 1.1.0' from 'nuget.org'.
    Retrieving package 'Microsoft.Win32.Primitives 4.3.0' from 'nuget.org'.
    Retrieving package 'NETStandard.Library 1.6.1' from 'nuget.org'.
    Retrieving package 'System.AppContext 4.3.0' from 'nuget.org'.
    ... [~20 more packages]
    Executing nuget actions took 41.31 sec

This doesn't happen on net46 framework.

    > nuget install FsCheck -Framework net46
    ...
    Retrieving package 'FsCheck 2.10.3' from 'nuget.org'.
    Retrieving package 'FSharp.Core 4.1.17' from 'nuget.org'.
    Retrieving package 'System.ValueTuple 4.3.0' from 'nuget.org'.
    ...
    Executing nuget actions took 1.66 sec

hickford:
This is the same problem as https://github.com/Microsoft/visualfsharp/issues/3451 for FSharp.Core , for which the solution was https://github.com/Microsoft/visualfsharp/pull/3452 "Basically FSharp.Core needs an empty dependency group for net45."

hickford:
As I understand, it would suffice to edit `FsCheck.nuspec`

    <dependencies>
      <group>
        <dependency id="FSharp.Core" version="4.1.0"></dependency>
      </group>
      <group targetFramework=".NETStandard1.6">
        <dependency id="NETStandard.Library" version="1.6.1" exclude="Build,Analyzers"></dependency>
        <dependency id="FSharp.Core" version="4.1.18" exclude="Build,Analyzers"></dependency>
      </group>
    </dependencies>

Adding an extra group

      <group targetFramework="net45">
        <dependency id="FSharp.Core" version="4.1.0"></dependency>
      </group>

kurtschelfthout:
Hm, okay. Not sure how to do that in the paket.tempate file though.

hickford:
[...]

PR #411, opened 2017-11-06
Add net45 dependency in nuspec.

kurtschelfthout:
See #410

### DIFF
--- a/src/FsCheck/paket.template
+++ b/src/FsCheck/paket.template
@@ -25,7 +25,7 @@ description
     Properties are written in F#, C# or VB. FsCheck helps you to define properties, observe the distribution of test data, and define test data generators. Crucially, when a property fails, FsCheck automatically displays a minimal counter example.
 dependencies
     FSharp.Core >= 4.1.0
-	framework: net45
+	framework: net452
 		FSharp.Core >= 4.1.0
 files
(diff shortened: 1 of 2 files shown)


hickford:
What nuspec(s) does this produce? 

Reading [build.cmd], I see it [merges two packages using mergenupkg](https://github.com/fscheck/FsCheck/blob/master/build.fsx#L268)

src/FsCheck/paket.template
```diff
@@ -25,6 +25,8 @@ description
     Properties are written in F#, C# or VB. FsCheck helps you to define properties, observe the distribution of test data, and define test data generators. Crucially, when a property fails, FsCheck automatically displays a minimal counter example.
 dependencies
     FSharp.Core >= 4.1.0
+	framework: net452
+		FSharp.Core >= 4.1.0
```
hickford: Try FSharp.Core 4.1.18 here? Earlier versions of FSharp.Core pick up extra packages https://github.com/Microsoft/visualfsharp/issues/3451

kurtschelfthout: Can't do that: 4.1.17 is the last FSharp.Core to have PCL builds.

hickford: Fair enough.

Issue #155, opened 2015-09-30
Cover FsCheck.Xunit with some automated tests


moodmosaic:
To save some time from manually testing, _maybe_ we could cover FsCheck.Xunit with automated tests:

| Parametric properties |
| :-- |
| PropertyAttribute |
| PropertyAttribute derived types |
| PropertyAttribute with Arbitraries |

| Non-parametric properties |
| :-- |
| PropertyAttribute |
| PropertyAttribute derived types |

Most of the above scenarios were used to test #86 manually.

kurtschelfthout:
In some sense this is already covered, in that FsCheck.Test uses FsCheck.Xunit to test FsCheck (yes, it makes my head hurt too).

What would be nice is some tests that specifically test the integration, i.e. passing test works, failing test works etc in all those cases. 

But I don't really know how to write those tests, i.e. how do I write a test (that passes) that runs an xunit test that fails and check the result? And can we check things like the visual studio integration works? Reliability - no false positives/negatives - is my main concern here.

moodmosaic:
>And can we check things like the visual studio integration works?

No, I think we can't test this automatically, because we would be relying on particular test-runners.

>In some sense this is already covered, in that FsCheck.Test uses FsCheck.Xunit to test FsCheck

Agreed on this one.

>What would be nice is some tests that specifically test the integration

I'd be quite interested in verifying `QuietOnSuccess` via automated test(s).

If such tests were in place, we'd have probably [avoided going back and forth](https://github.com/fscheck/FsCheck/pull/145) on that one. Then, we can probably extrapolate for other properties as well. But to be honest I don't really know how to write those tests either.

kurtschelfthout:
I know of one way to write them - by running the console runner in a separate process and check the outputs - but that just seems like asking for trouble.

ploeh:
Perhaps... FsCheck could be written as a library of side-effect-free queries, reporting status as return values instead of writing to the console :stuck_out_tongue_winking_eye: 

If so, it'd be possible to test the return values of those queries.

At the boundary, FsCheck could still write those return values to the console.

ploeh:
Bantering aside, there's quite a bit of untested logic in PropertyAttribute.fs, and it might be a good idea to refactor it so that the implementation is as pure and testable as possible, leaving only the glue to xUnit.net untested.

I'll be happy to look into that whenever I have a bit of time, but I'll need to first understand how the xUnit.net 2 extensibility API works...

[...]

PR #414, opened 2017-12-09
Add support for XUnit's BeforeAfterTestAttribute

willsb:
The code here is based on what [`XUnitTestInvoker`](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs) does to process those attributes and takes advantage of the `XunitTestCaseRunner` for discovering the attributes.

### DIFF
--- a/src/FsCheck.Xunit/PropertyAttribute.fs
+++ b/src/FsCheck.Xunit/PropertyAttribute.fs
@@ -1,6 +1,9 @@
 ﻿namespace FsCheck.Xunit
 
 open System
+open System.Linq;
+open System.Reflection
+open System.Threading
 open System.Threading.Tasks
 
 open FsCheck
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Just checking, but what makes me feel a bit uneasy is that we are essentially duplicating code from xunit at least to a certain extent. This is independent from this PR, but is there perhaps a way to hook into xunit in such a way that a property test works exactly like a Fact for all the setup/tear down stuff without duplicating xunit code here?

Or did I misunderstand something?

willsb:
We could implement the whole chain up to `PropertyTestRunner` (`PropertyTestCase ` -> `PropertyTestCaseRunner ` -> `PropertyTestRunner` -> `PropertyTestInvoker`). I initially didn't follow this approach because the code only went as far as to implementing `PropertyTestCase`, so I thought the project's wouldn't want to leverage XUnit's way of implementing tests for some reason, so I only implemented what was absolutely needed. I'll refactor this to do things the XUnit way

willsb:
After further investigating this, here's what I found out:

- The codebase already copies some code from xUnit (namely the order of the calls to `messageBus.QueueMessage` and [the creation of the `testClassInstance`](https://github.com/fscheck/FsCheck/blob/master/src/FsCheck.Xunit/PropertyAttribute.fs#L195-L202)
- Extending `XUnitTestInvoker` is cumbersome because [`RunAsync `](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs#L187) is not virtual and [`CallTestMethod`](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs#L149) does not suite our needs without some refactor

With that in mind, the current point of extension the best possible without a bigger refactor. It uses xUnit for most things and keeping boilerplate at a minimum. 

FWIW, if one is interested in reducing the amount of reuse, a good point of extension would be [`XUnitTestRunner.InvokeTestAsync`](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestRunner.cs#L52-L76). A lot of what I mention to be duplicate would be removed if we extended from that point. That would, however, require a refactoring and integration of a custom `XUnitTestInvoker`, something that, IMO, should not be part of this PR's scope.

ploeh:
I agree with @kurtschelfthout that I'd be concerned duplicating even more code from xUnit.net here. We've previously seen defects in _FsCheck.Xunit_ making it into production because we have no test coverage. We also [tried adding tests](https://github.com/fscheck/FsCheck/issues/155), but so far, no-one has been able to do that.

> Extending `XUnitTestInvoker` is cumbersome because `RunAsync` is not virtual and `CallTestMethod` does not suite our needs without some refactor

Would it be possible to work with xUnit.net to add the required extensibility points there, instead of duplicating effort here?

Ideally, the duplication that we _already_ have should be folded back into xUnit.net, if possible.

kurtschelfthout:
@willsb Thanks a lot for the investigation.

> Would it be possible to work with xUnit.net to add the required extensibility points there, instead of duplicating effort here?

Sounds like a good idea - however sounds like we are not using the extension points that they have:

> FWIW, if one is interested in reducing the amount of reuse, a good point of extension would be XUnitTestRunner.InvokeTestAsync. A lot of what I mention to be duplicate would be removed if we extended from that point. 

So they might just point us that way, which is also reasonable.

Other than the broader question, any thoughts on what to do with this PR right now? Personally I am inclined to accept it to unblock @willsb but then make it clear in the issues that we intend to refactor as above and so won't be accepting more of these kind of copy-pastes from xunit.

ploeh:
For a long time, I've been thinking about bringing this issue to the attention of the xUnit.net maintainers, in order to discuss how this can be addressed, but I haven't prioritised it yet. While xUnit.net has a nice feature for data-driven tests, in it's current incarnation I'm not sure it's flexible enough to enable us to hook into it.

Specifically, xUnit.net has a `[<Theory>]` attribute that third parties can use to attach data sources. This is, for example, the mechanism that [AutoFixture](https://github.com/AutoFixture/AutoFixture) uses for its [AutoDataAttribute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit2/AutoDataAttribute.cs). The problem is, however, that this extensibility point currently isn't useful for FsCheck.

The reason is that property-based tests potentially go through two phases:

1. Data generation
2. Shrinking

The `[<Theory>]` extensibility point only enables third parties to hook into the data generation phase. We could trivially hook into this in order to generate (by default) 100 test cases, but AFAIK, there's no extensibility point that enables us to, say, stop execution if test case no. 32 falsifies the property. Additionally, I don't think there's any way to hook into a failing test case and start the shrinking phase.

It'd be really nice if xUnit.net offered extensibility points that would enable these capabilities, because that would mean that the rest of xUnit.net's API would just automatically work, including the `[<BeforeAfterTest>]` attribute.

FsCheck isn't the only library with this need. [Hedgehog](https://github.com/hedgehogqa/fsharp-hedgehog) has the exact same need, so perhaps the xUnit.net maintainers would be interested in working with us regarding adding such capabilities.

---

> Other than the broader question, any thoughts on what to do with this PR right now?

I think we ought to close it for now in order to investigate other options first. I'm already concerned about the maintainability level of the current FsCheck.Xunit code base (particularly given that we don't have tests). Adding more untested code to it would be the straw that breaks the camel's back, I'm afraid.

Another way to look at is this: Does adding support for `[<BeforeAfterTest>]` warrant the extra complexity? After all, use of this (and derived) attribute(s) strongly implies that mutable state is involved, so again, by adding support for something like that, we could be painting ourselves into a new corner that prevents running properties in parallel.

kurtschelfthout:
Ok, I think that's a fair assessment. Closing. Will also open a new issue to track xunit integration woes.

PR #416, opened 2017-12-22
[WIP] fix sdk2

enricosada:
- fix nuspec with dotnet-mergenupkg v2
- convert to .net core sdk2 maintaining current supported fw


kurtschelfthout:
What happened here? :) some of those bits look intriguing.

enricosada:
@kurtschelfthout np, i'll reopen soon :D

PR #417, opened 2017-12-25
remove fallback group from nupkg

enricosada:
bump dotnet-mergenupkg to v2.2.
this remove fallback group from nupkg

cleanup for .net core sdk 2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       dist: trusty
       sudo: required
       mono: latest
-      dotnet: 2.0.0
+      dotnet: 2.0.3
 
 script:
   - ./build.sh RunTests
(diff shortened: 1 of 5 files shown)


enricosada:
@kurtschelfthout this replace #416 .

i left other stuff, like bumps xunit, netstandard2.0 out, because with these two test fails, i'll send it later.

this pr fix the issue with fallback group, so will be nice to have a pkg. if you need also for v3 branch, just bump dotnet-mergenupkg to `2.*`. i can send a pr there too if needed

kurtschelfthout:
Awesome, thanks! I'll take care of the v3 branch. And releasing.

kurtschelfthout:
Released in v2.10.5. Also merged to v3 branch.

PR #424, opened 2018-01-21
Propattr failure msgs

adj123:
Proposed changes to get failure info showing when calling from NUnit

PR #425, opened 2018-01-21
Implies extension method

adj123:
Also, another thing I've found is that calling FsCheck via NUnit in C# when using preconditions is a bit syntactically awkward. (Specifically if I don't want the property body to execute when preconditions aren't satisfied.) For a property `someBoolean` parameterised by an `int` `arg` subject to a precondition `somePrecondition`, the test has to be expressed like below:

```
[Property]
public static bool SomeProperty(int arg)
{
    Func<bool> property = () => someBoolean;
    return property.When(somePrecondition);
}
```

The `property` delegate can't really be tidied up; the compiler won't accept a local function in its place, and if it's inlined into the `return` statement then it needs to be explicitly cast to `Func<bool>` still for the extension method to apply. I've found that adding an extension method which acts the opposite way round to `When` allows the delegate to be implicitly cast so the above can be tidied up to:

```
[Property]
public static bool SomeProperty(int arg) => 
    somePrecondition.Implies(() => someBoolean);
```

I use this wrapper in my own testing library but since it seems universally useful I suggest it as part of the FsCheck library itself (though maybe it should specifically be in the FsCheck.NUnit project instead, I don't know whether it's useful for other clients).


kurtschelfthout:
Same comment as #424 unfortunately...

Also, should it be called `Then`?

adj123:
Sure, I'm easy - personally I prefer `Implies` for the parallels with the F# `==>` operator, since it's the C# analogue of that, but I'm happy to change it.

kurtschelfthout:
> personally I prefer Implies for the parallels with the F# ==> operator

Yes, fair point.

PR #426, opened 2018-01-21
Amended NUnit PropertyAttribute to show info messages on failure

adj123:
This looks better :)


kurtschelfthout:
Released in 2.10.7.

PR #427, opened 2018-01-21
Added Implies extension method

adj123:
Fixed from the previous PR to go from master


src/FsCheck/PropExtensions.fs
```diff
@@ -40,7 +40,7 @@ type PropertyExtensions =
     ///Conditional property combinator. Resulting property holds if the property holds when the condition does.
     [<Extension>]
     static member Implies(condition, property:Func<bool>) = 
-        property.When(condition)
```
adj123: Think the appveyor build must be targeting an older version of F# since it didn't seem to like this but it built locally on my machine ok - I updated it not to use the extension method syntax and now it seems to run fine.

kurtschelfthout:
Released in 2.10.7.

PR #435, opened 2018-04-21
Add link to Expecto.FsCheck

haf:


kurtschelfthout:
There is already a link to expecto in the docs https://fscheck.github.io/FsCheck/QuickStart.html#Integration-with-Expecto I think that's enough. That space in the Readme is for the releases that are done specifically from this repo.

haf:
People make decisions based on what docs are available not on where releases are made.

PR #437, opened 2018-05-09
Feature/fix sublist of

MNie:
I spot a problem with FsCheck when I'm using Gen.SublistOf with an argument which contains only a single element. I think in this corner scenario we should generate arrays with this single element and empty arrays. Right now only empty arrays would be generated.
I'm not 100% sure about the unit test if it suits how it should be written so any suggestion would be great!

CC: @kurtschelfthout

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -533,8 +533,7 @@ module Gen =
     let subListOf l =
         let elems = Array.ofSeq l
         gen {// Generate indices into the array (up to the number of elements)
-             let maxNumberOfElements = if elems.Length = 1 then 1 else elems.Length-1
-             let! size = choose(0, maxNumberOfElements)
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
I agree - a sublist of a list can also be the whole list. But you don't need to special case for the 1-element list. Just remove the `-1` from the upper bound of the `let! size = choose(0, elems.Length-1)` instead.

kurtschelfthout:
OK. I agree with the fix, but the test is cumbersome, and fails on AppVeyor.

Can you do something like `Gen.subListOf [1] |> Gen.sample 100 100` and check that the resulting list of lists contains a `[1]` list.

MNie:
Fix'd :)

kurtschelfthout:
Thanks very much!

kurtschelfthout:
This is now released in 2.10.10.

PR #441, opened 2018-06-14
Add DateTimeKind and milliseconds to default generator for DateTime.

mjmckp:
Also, add generator for `DoNotSize DateTime`.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -197,8 +197,7 @@ module Arb =
                         |> Seq.takeWhile ((|>|) n) }
         |> Seq.distinct
 
-    [<EditorBrowsable(EditorBrowsableState.Never)>]
-    let inline shrinkDate (d:DateTime) =
+    let internal shrinkDate (d:DateTime) =
         if d.Kind <> DateTimeKind.Unspecified then
             seq { yield DateTime.SpecifyKind(d, DateTimeKind.Unspecified) }
         elif d.Millisecond <> 0 then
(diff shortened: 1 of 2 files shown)


src/FsCheck/Arbitrary.fs
```diff
@@ -197,6 +197,21 @@ module Arb =
                         |> Seq.takeWhile ((|>|) n) }
         |> Seq.distinct
 
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
```
kurtschelfthout: Make it internal?

tests/FsCheck.Test/Arbitrary.fs
```diff
@@ -260,9 +260,11 @@ module Arbitrary =
     [<Property>]
     let ``DateTime shrinks seconds, minutes and hours`` (value:DateTime) =
         shrink value
-        |> Seq.forall (fun v -> v.Second = 0 
-                                 || (v.Second = 0 && v.Minute = 0) 
-                                 || (v.Second = 0 && v.Minute = 0 && v.Hour = 0) )
+        |> Seq.forall (fun v -> v.Kind = DateTimeKind.Unspecified   // TODO: all the cases below are redundant?
```
kurtschelfthout: Indeed they are. A stronger check would be to check that if the hours are 0, then also minutes, seconds & kind are 0 resp. unspecified, and so on.

mjmckp: Not sure that would work, since it's possible that the original value happened to have `Hour` equal to zero.  I don't think there is much that can be asserted here except `v.Kind = DateTimeKind.Unspecified`, given the implementation of the shrinker.

kurtschelfthout:
Awesome, thanks. 

I'll release this tonight or over the weekend.

Issue #422, opened 2018-01-16
Support for struct tuples?


eiriktsarpalis:
It seems that the latest prerelease of FsCheck is lacking support for struct tuple generation. I could provide an implementation for this, if I find the time.

kurtschelfthout:
Indeed, good point. I choose to interpret that as you're working on it until I hear otherwise ;-p

kurtschelfthout:
Ah, that's unfortunate. If that would work, saves a dependency on ValueTuple nuget in some cases, right? Should we just add an explicit Arbitrary instance for now anyway?

kurtschelfthout:
Just noticed this actually works out of the box as long as you use an updated FSharp.Core with the fix mentioned above (I tested with NuGet FSharp.Core version 4.5.0):

```fsharp
    Arb.generate<struct (int*bool)> |> Gen.sample 10 10 |> printfn "%A"
    Arb.shrink (struct (15,1)) |> Seq.toArray |> printfn "%A"
```

Both works. So I'm going to close PR #442  - don't think it's worth getting this to work for older FSharp.Core versions, and besides the redirect and packaging issues for FsCheck itself are giving me headaches.

I consider this issue as fixed.

PR #442, opened 2018-06-16
Support ValueTuple generation and shrinking. (#422)

kurtschelfthout:
Works on  .NET Framework, builds on .NET Core but doesn't work at
runtime because it can't find the assembly.


kurtschelfthout:
Closing as working with an updated FSharp.Core (tested with 4.5.0) makes this work out of the box. See issue #422

PR #445, opened 2018-08-11
[Proposal] Add XML serializable value as generated type

stijnmoreels:
The genereted type is a string that can be used as a XML value inside a document.

Several times now, I had to deal with generated values that should be serializable as an XML and I wanted to have a broader range that just using GUIDs in all places.

Just an idea, love to hear your feedback!

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -55,8 +55,8 @@ type StringNoNulls = StringNoNulls of string with
     override x.ToString() = x.Get
 
 ///Represents a string that can be serializable as a XML value.
-type XmlValue = XmlValue of string with
-    member x.Get = match x with XmlValue v -> v
+type XmlEncodedString = XmlEncodedString of string with
+    member x.Get = match x with XmlEncodedString v -> v
     override x.ToString() = x.Get
 
 ///Represents an integer interval.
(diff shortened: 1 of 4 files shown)


kurtschelfthout:
It doesn't hurt I suppose :)

Could the name `XmlValue` be somewhat improved perhaps? `XmlEncodedString`? Don't know the XML lingo that well.

ploeh:
Or just `XmlString`?

ploeh:
Or perhaps `XmlContents`

As far as I can see, this generator just creates a single string that's valid as the contents of an XML element. Is that correct?

At first glance, I understood the name `XmlValue` as something that could, potentially, be an entire XML document, but if I understand this pull request correctly, it's not that.

I could be mistaken, though...

stijnmoreels:
@ploeh Yes, its sole purpose it to have a string that can be used _inside_ a XML document. Serializable as a inner text value, so to speak.

I didn't find a good name for it, and thought about the `XmlNode` class itself, and the way it had a `Value` property. But, yeah, without it; it sounds like a complete XML document.

The `XmlEncodedString` proposal is (I think) the closest to what it is I suppose.

stijnmoreels:
Everyone agreed on renaming `XmlValue` > `XmlEncodedString`?

kurtschelfthout:
> Everyone agreed on renaming XmlValue > XmlEncodedString?

Yes, I think so. (At least I am.)

ploeh:
Sounds good to me 👍

stijnmoreels:
Ok, I think for me this is ready to merge.

kurtschelfthout:
Thanks!

PR #454, opened 2018-10-13
FEAT: add missing csharp extensions

stijnmoreels:
- zip functionality was missing from the csharp extensions
- when creating custom generators in csharp, we often require to check
for `null`, meaning we have to generate an additional `null` value so I
often use the `OrNull` extension method to simply specify this value.

(this last one is just an idea, if it doesn't make sense please say so and I remove it from this PR)

These extensions aren't actually tested because it just adds a facade on top of the fsharp operators, is that correct?

Thanks!


kurtschelfthout:
I think this is fine as is, re: testing sometimes I just add an example in one of the C# example projects just to check things are working as expected. Feel free to follow up with another PR if you think that is useful.

PR #455, opened 2018-10-14
FEAT: introduce `traverse` from the existing `sequence` functions

stijnmoreels:
We already had `sequence` functions for sequences, lists and arrays; but we could also define `traverse` functions for it.

Again (!) just and idea during writing custom generators.
Did I correctly introduced these `traverse` functions?

Thanks!


kurtschelfthout:
Yes, Looks like they’re correctly implemented. 

I’ve been thinking we should use the F# naming where appropriate - `traverse` is pretty similar to `Seq.collect` so we could use `collect` as the root for the F# names, as this may be more familiar to people. 

For C# I’m not sure - is there anything in Linq that is similar? `Aggregate` maybe?

stijnmoreels:
> I’ve been thinking we should use the F# naming where appropriate - `traverse` is pretty similar to `Seq.collect` so we could use `collect` as the root for the F# names, as this may be more familiar to people.

Yes, you could be right there. `traverse` is maybe a bit too much _Haskelly_. I'll rename this.
 
> For C# I’m not sure - is there anything in Linq that is similar? `Aggregate` maybe?

There's something like `Task.WhenAll` which is kind-of the same as the `sequence`; maybe is `Collect` also a good name to be used for C#.

stijnmoreels:
Hmm, that about the `SelectMany` was wrong. I rename it to `Collect`.

ploeh:
I've never been able to figure out from where F# takes its terminology, but it's certainly not from Haskell, and neither is it from C#. Could the choice of function names come from OCaml?

Does OCaml have a function that's equivalent to Haskell's `traverse`, and if so, what's it called?

stijnmoreels:
> Does OCaml have a function that's equivalent to Haskell's `traverse`, and if so, what's it called?

From the OCaml language reference:
image

ploeh:
I don't know... I don't know OCaml, but from how I _guess_ that I have to read that type, it doesn't quite correspond to Haskell's `traverse`:

```haskell
traverse :: Applicative f => (a -> f b) -> t a -> f (t b)
```

kurtschelfthout:
@ploeh From Don? :)

OCaml has `concat_map` https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/#val-concat_map looks like.

kurtschelfthout:
Cheers!

PR #456, opened 2018-10-16
Fix timezone generator for DateTimeOffset

eugene-g:
Most of the time zones on land are offset from Coordinated Universal Time (UTC) by a whole number of hours (UTC−12 to UTC+14 (c) [Wikipedia](https://en.wikipedia.org/wiki/Time_zone)

PR #458, opened 2018-10-17
Fix timezone generator for DateTimeOffset

eugene-g:
Recreating #456


kurtschelfthout:
Thank you very much, and sorry for the extra hassle.

PR #460, opened 2018-11-07
FEAT: add 'NonWhiteSpaceString' to the default generated types

stijnmoreels:
Since we also have a `String.IsNullOrWhiteSpace` and this is used quite
a lot for guard assertions and validations; it seems logical to also
generate strings without whitespace characters.

Seems reasonable?

Thanks.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -54,6 +54,11 @@ type StringNoNulls = StringNoNulls of string with
     member x.Get = match x with StringNoNulls r -> r
     override x.ToString() = x.Get
 
+// Represents a string that is not null, empty or consists only of white-space characters and does not contain any null characters ('\000')
+type NonWhiteSpaceString = NonWhiteSpaceString of string with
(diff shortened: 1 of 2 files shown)


ploeh:
[That method is evil and should be ignored](http://blog.ploeh.dk/2014/11/18/the-isnullorwhitespace-trap).

stijnmoreels:
> [That method is evil and should be ignored](http://blog.ploeh.dk/2014/11/18/the-isnullorwhitespace-trap).

I agree! Although, when deserializing to domain models; they sometimes could be seen as "the same".
But thanks @ploeh for the comment! If you rather would see that this PR is abbandoned; it's fine by me.

ploeh:
> If you rather would see that this PR is abbandoned

No, I'll let @kurtschelfthout make that decision.

I now realise that my comment was insensitive. I apologise 😳

stijnmoreels:
> I now realise that my comment was insensitive. I apologise 😳

No problem ;)

kurtschelfthout:
I think it’s a reasonable addition, and largely independent of the value or evilness :) of the `String` method?

PR #461, opened 2018-12-04
Add DoctestInteractive to users.

tomakita:
I made a project that uses FsCheck, so I thought I'd add it to the Users page.  Let me know if I did this incorrectly!


kurtschelfthout:
great, thank you!

PR #463, opened 2019-01-11
Shrink DTO classes generated using reflection based generators

czan:
When using FsCheck from C# we noticed that it didn't shrink types when we used a reflective generator. That was annoying, so here's a patch to fix it.

### DIFF
--- a/src/FsCheck/Reflect.fs
+++ b/src/FsCheck/Reflect.fs
@@ -101,6 +101,16 @@ module internal Reflect =
         let f     = l.Compile ()
         f.Invoke
 
+    let getCSharpDtoReader (recordType: Type) =
+        if isCSharpDtoType recordType then
+            let properties = getProperties recordType
+                             |> Seq.filter (fun p -> p.CanWrite)
+                             |> Seq.map (fun p -> p.GetValue)
+                             |> Seq.toArray
+            let lookup o = Array.map (fun f -> f o) properties
+            lookup
+        else
+            failwithf "The input type must be a DTO class. Got %A" recordType
 
     /// Returns the case name, type, and functions that will construct a constructor and a reader of a union type respectively
     let getUnionCases unionType : (string * (int * System.Type list * (obj[] -> obj) * (obj -> obj[]))) list = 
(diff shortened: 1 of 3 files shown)


kurtschelfthout:
Excellent, thanks!

Can you please add tests for the new functionality here: https://github.com/fscheck/FsCheck/blob/master/tests/FsCheck.Test/Arbitrary.fs#L581

czan:
Apologies for the constant force pushes and rebuilds.

I would like to write a working test, but I can't actually get FsCheck to build on my computer. To test my original patch I had to send it to a coworker's computer where he built it and ran our tests.

I think I have the test close to working, but without being able to run it locally the feedback loop of force pushing is too slow for me to keep probing, given I don't really know what I'm doing.

kurtschelfthout:
> I can't actually get FsCheck to build on my computer. 

Well that's a sad situation! Can you run `build.cmd` (on Windows) or `build.sh` (on *Nix) - both in the root folder of the repo - and send the output?

kurtschelfthout:
And don't worry about the force pushing. If we can't get it to build on your machine, I can take a look at fixing this up tonight or this weekend.

czan:
Sure! I cleared everything out and tried again and it still failed. I've put the information that I thought was necessary here: https://gist.github.com/czan/804b4fd5c62804f3f40a7e11634dc9e4

If there's anything else that would be helpful, let me know. Would you like me to open an issue for this, rather than trying to solve my issues on this PR?

kurtschelfthout:
I'm not really sure I can help much, though that error message does look annoyingly familiar. I think I had that when I was switching between .net framework "old" build system (which may be mono in your case) and the "new" build system. There's a bunch of NuGet and build caches that can confuse things. Perhaps removing your local repository altogether and cloning it again would help?

kurtschelfthout:
Maybe this is interesting. Your `usr/bin/msbuild` seems to be the mono build, while on the Travis build it used the dotnet msbuild. So perhaps it's some sort of path/msbuild issue? Not quite sure how to address it though, besides somehow getting .NET Core's msbuild in /usr/bin somehow. But that may break other stuff on your machine?

PR #469, opened 2019-02-28
Default arbitrary for Numerics.Complex

Kharacternyk:
Complex values are shrunk by removing imaginary part and ordinary shrinking both parts.

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -487,6 +487,21 @@ module Arb =
                 }
             fromGenShrink (genDecimal, shrinkDecimal)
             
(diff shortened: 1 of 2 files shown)


Kharacternyk:
Well, something isn't right here. It perfectly compiles on my machine and the message about undefined Complex type is weird. Am I missing something?

kurtschelfthout:
I think that it is failing in the .NET Standard 1.6 build because presumably that type is not defined there. Some compiler directives may be necessary to exclude the code from that build. There are other examples of that in Arbitrary.fs.

Kharacternyk:
Ok, I'll add some #if directives later

Kharacternyk:
This branch is so messy for such little improvement. I'll rewrite all of this in the new branch with single commit. Thanks for help with multitargeting!

PR #470, opened 2019-03-01
Default arbitrary for Numerics.Complex

Kharacternyk:
Clean-upped version of #469

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -486,7 +486,27 @@ module Arb =
                     if truncated |<| d then yield truncated
                 }
             fromGenShrink (genDecimal, shrinkDecimal)
-            
+
+#if !NETSTANDARD1_0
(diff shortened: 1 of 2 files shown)


kurtschelfthout:
@Kharacternyk could you take a look at the test failure on travis?

 FsCheck.Test.Arbitrary.Complex shrinking produce values without duplicates and at least one without imaginary part [FAIL]
      FsCheck.Xunit.PropertyFailedException : 
      Falsifiable, after 14 tests (0 shrinks) (StdGen (558543933,296566765)):
      Original:
      (NaN, Infinity)

Kharacternyk:
Well, I hadn't caught it myself. It seems, I've used a <> operator in filtering, but Seq.distinct uses Equals, so shrinking produces  {(Nan, 0); (Nan; 0)}, that isn't distinct. I'll fix it

Kharacternyk:
To be exact, Seq.distinct uses hashes, and hashes of (Nan; 0) and (Nan; 0) are equal, while (Nan; 0) <> (Nan; 0) returns true.

PR #471, opened 2019-03-02
Implemented shrinker for the Interval type

Kharacternyk:


src/FsCheck/Arbitrary.fs
```diff
@@ -846,9 +846,14 @@ module Arb =
...
-            gen { let! start,offset = Gen.two generate
-                  return Interval (abs start,abs start+abs offset) }//TODO: shrinker
-            |> fromGen
+            let generator = 
+                  gen { let! start,offset = Gen.two generate
+                  return Interval (abs start,abs start+abs offset) }
+            let shrinker (i : Interval) =
+                (i.Left, i.Right - i.Left)
+                |> shrink
+                |> Seq.map (fun (right, length) -> Interval(right, right + length))
```
kurtschelfthout: nit: this should be `fun (left, length) ->` etc

kurtschelfthout: and actually, while we're at it, might as well make the terminology consistent between generator and shrinker. Either `left` `right` and `length` or `start` and `offset`. Would prefer the former.

tests/FsCheck.Test/Arbitrary.fs
```diff
@@ -372,6 +372,16 @@ module Arbitrary =
...
         shrink value |> Seq.forall (fun (NegativeInt v) -> v < 0)
 
+    [<Property>]
+    let ``Interval after shrinking is smaller than origin`` (i : Interval) =
+        let shrunk = shrink i
+        let isWithoutDuplicates = (shrunk |> Seq.distinct |> Seq.length) = (Seq.length shrunk)
+        let isSmaller = 
+            shrunk
+            |> Seq.map (fun i -> match i with Interval (right, left) -> (right, left))
+            |> Seq.forall (fun (right, left) -> (right <= i.Right) && (right - left <= i.Right - i.Left))
```
kurtschelfthout: Here also in both lines `right` and `left` are the other way around.

Kharacternyk:
Huh, I'm not native speaker, so these obvious mistakes aren't such obvious for me. I'll fix it later. Also, I'll replace length with offset

Kharacternyk:
Here it is. I'll be more careful in the future

Kharacternyk:
Well, instead of merging I should have done rebase. Again, I'll create new pull request with single commit

PR #475, opened 2019-03-04
Fixed Complex shrinker

Kharacternyk:
Duplicates might occur, because NaNs weren't properly compared via Equals. See #470


Kharacternyk:
Though the shrunk sequence produced duplicates (this PR fixes it), it seems that Seq.distinct didn't remove NaNs too (I tried running failed stdGen and it passed even without any changes to code). Thus this PR may not make test passed on travis. I hope it won't be, but if so, I don't know how to fix thing that I can't even reproduce

kurtschelfthout:
Thanks much!

Kharacternyk:
Please let me know if it still fails. I'll continue trying solution

kurtschelfthout:
Looks like it is working now.

PR #476, opened 2019-03-05
Added default arbitrary for ConsoleKeyInfo

Kharacternyk:
System.Console.ReadKey() returns an instance of ConsoleKeyInfo type. Default arbitrary may be useful for testing some console input scenarios


kurtschelfthout:
Wasn't the first type I expected but ok :)

PR #481, opened 2019-03-11
Always provide the own implementation of the printf operator in Runner.fs

Kharacternyk:
I was using it for a while as unstaged change. Though appveyor and travis are happy with it as it is, compiler on my machine complains about missing printf definition when builds for .NET Standard 1.6 and 2.0. If this PR isn't acceptible, are there any thoughts about origin of this problem?

### DIFF
--- a/src/FsCheck/Runner.fs
+++ b/src/FsCheck/Runner.fs
@@ -245,10 +245,8 @@ module Runner =
     let onShrinkToString args =
         sprintf "shrink:%s%s%s" newline (argumentsToString args) newline
 
-#if NETSTANDARD1_0
     let internal printf fmt = 
         Printf.kprintf Diagnostics.Debug.WriteLine fmt
-#endif
     
     ///A runner that prints results to the standard output.
     let consoleRunner =


kurtschelfthout:
I'm not sure I understand. `printf` is not available in FSharp.Core builds for portable libraries, which we're using instead of a proper .netstandard 1.0 build of FSharp.Core (which does not exist).

If a `printf` exists in FSharp.Core we should use that, and not this workaround, hence the existing compiler directive.

Kharacternyk:
I'll send exact compiler message tomorrow. In short: "haven't found printf definition, try these: Printf(module), printfn...". It occures when I try to build for standard 1.6 or 2.0 (no problem with 1.0 or framework451). But of course it may be my local problem. I'm sorry for such confusing PR, maybe issue would be better

kurtschelfthout:
Ok, but that's not what I meant, I don't think this change is desirable. We only need the definition if it's not available in FSharp.Core. 

What are you trying to achieve?

Kharacternyk:
Without this change I just can't build solution. I thought that if this change doesn't break anything it will be ok to merge it, so people with same problem will be happy. But it seems I'm the only one with this problem.

Kharacternyk:
I'll just left it as my local change

kurtschelfthout:
Oh I see. That's weird, yes I certainly haven't heard of this problem before. Maybe some issue with an FSharp.Core version?

Kharacternyk:
Huh, Visual Studio tells me that .NET standard 1.6 and 2.0 builds use version 4.0.0.1; in folder "packages" (generated after I executed "paket restore") there is version 4.2.3; and in my local user nuget cache there are three versions: 4.0.0.1, 4.2.3 and 4.5.2.

kurtschelfthout:
maybe try running `dotnet restore`? Or run build.cmd which will do that.

Kharacternyk:
Didn't help. `error FS0039: The value or constructor 'printf' is not defined. Maybe you want one of the following:   Printf   fprintf   sprintf   fprintfn   PrintfFormat`

kurtschelfthout:
Sorry I don't know...do you have an updated version of the compiler/visual studio if that's what you're using?

Kharacternyk:
Yep, VS 2017 community, latest update. Well, nevermind :) It isn't serious issue. I'll either build for framework, or just keep it ustaged. Thanks much for trying to help!

Kharacternyk:
I've discovered recently that when I build while being on fscheck3 branch all is ok, though these lines haven't changed. So weird.

kurtschelfthout:
It does indicate there is something out of whack with packages sounds like...that stuff can be frustrating. At the time I was upgrading from old to new project format I blew away my entire repo a few times, as well as the nuget cache, just to make sure I started with a clean slate.

Kharacternyk:
I've updated to VS2019 recently and this issue has disappeared. But... now I have `"fsc.exe" exited with code -1073741571.` when I try to build for .net standard 1.0 or 1.6 within the VS. So I've tried to use dotnet CLI and it has built FsCheck for all targets successfuly. Maybe this whole story was just a bug inside the VS.

PR #482, opened 2019-03-13
[Sketch] Add arbitrary for System.Type

Kharacternyk:
Generator can produce non-generic types and generic types with non-generic typeparams. Shrinking is not supported yet. Sample:
```
Ok, passed 100 tests.
9% System.Text.StringBuilder.
8% System.String.
7% System.Int32.
7% System.Byte.
5% System.Enum.
5% System.Array.
4% System.Object.
3% System.Threading.Tasks.Task`1[System.Text.StringBuilder].
3% System.Threading.Tasks.Task`1[System.String].
3% System.Threading.Tasks.Task`1[System.Byte].
3% System.Collections.Generic.List`1[System.Array].
2% System.Threading.Tasks.Task`1[System.Object].
2% System.Threading.Tasks.Task`1[System.Enum].
2% System.Threading.Tasks.Task`1[System.Array].
2% System.Collections.Generic.List`1[System.Type].
2% System.Collections.Generic.List`1[System.Object].
2% System.Collections.Generic.KeyValuePair`2[System.Int32,System.Type].
2% System.Collections.Generic.Dictionary`2[System.Text.StringBuilder,System.Object].
2% System.Collections.Generic.Dictionary`2[System.Text.StringBuilder,System.Int32].
2% System.Collections.Generic.Dictionary`2[System.Enum,System.Text.StringBuilder].
1% System.Type.
```
Should we produce "nested" generic types? Thoughts about shrinker?

### DIFF
--- a/src/FsCheck/Arbitrary.fs
+++ b/src/FsCheck/Arbitrary.fs
@@ -1204,6 +1204,29 @@ module Arb =
         static member BigInt() =
             Default.Int32()
             |> convert bigint int
+        
+        ///Arbitrary instance for System.Type. Do not support shrinking
+        static member Type() =
+            let nonGeneric = Gen.elements [
+                typeof<int32>; typeof<byte>; //Primitives
+                typeof<string>; //Immutable reference type, sealed 
+                typeof<Enum>; //Derived from ValueType, but not value type itself
+                typeof<System.Text.StringBuilder>; //Mutable reference type
+                typeof<Type>;
+                typeof<Array>;
+                typeof<Object>
+            ]
+            let rec genericHelper degree (types: Type list) =
+                let genericBuilder (types : Type list) (template : Type) =
+                    template.GetGenericTypeDefinition().MakeGenericType (List.toArray types)
+                Gen.map2 
+                    genericBuilder
+                    (Gen.listOfLength degree nonGeneric)
+                    (Gen.elements types)                     
+            let genericOne = genericHelper 1 [typeof<List<_>>; typeof<System.Threading.Tasks.Task<_>>]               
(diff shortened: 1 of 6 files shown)


Kharacternyk:
By the way, does defining types like "GenericType", "ReferenceType" etc. make sense?

kurtschelfthout:
Interesting!

I wonder how useful a completely arbitrary `Type` generator is. I expect the point of this would be to test some reflection code. But in that case there are typically some constraints on the types (e.g. implement a particular interface, derive from a particular class, have properties with certain names, etc). With that, for this we'd need a much more configuration generator/shrinker, possibly an entirely separate API. 

I think this is a lot of work!

Kharacternyk:
Yeah, I see. Maybe it shouldn't be the default arbitrary, but something like "constructors" to easily generate types with some constraints. I'll think about it.

kurtschelfthout:
> but something like "constructors" to easily generate types with some constraints.

Yes exactly. I played with this a little bit a long time ago, thinking about how to generate types that implement a given interface, have a particular attribute, have a particular method. It's a large surface and you quickly get into mock-like territory. (i.e. if you generate types that implement an interface, then what should the implementation be? Default generator for the return type?). It would also involve some runtime code generation, which is not the easiest to get right.

Frankly I think this could be an entirely separate project in its own right, like an extension to FsCheck.

Kharacternyk:
I think I'm not ready to implement such a big piece of API by myself yet. Closing it, maybe I'll open another PR when I have a sketch of entire type-constructing-on-the-fly API. I don't want to confuse you with my WIP commits that may be reverted next day.

PR #489, opened 2019-03-23
Use all 28 significant digits in default Decimal gen.

Kharacternyk:
Previous implementation used only 9 significant digits.


Kharacternyk:
.NET Standard 1.0 doesn't have BigInteger type, so it uses FSharp.Core ones, that doesn't have conversion to Decimal... Well, it requires some more efforts.

Kharacternyk:
It isn't much elegant, but other solutions are too complex for such specific using case.

PR #490, opened 2019-03-24
Get rid of allocating arrays in stdFloatGen implementation

Kharacternyk:


src/FsCheck/Arbitrary.fs
```diff
@@ -438,13 +438,10 @@ module Arb =
...
-            let toFloat bytes =
-                BitConverter.ToUInt64(bytes, 0)
-                |> fun n -> (float (n >>> 11)) * (1.0 / float (1UL <<< 53))
-                |> BitConverter.GetBytes
-                |> fun b -> BitConverter.ToDouble(b, 0)
-            Gen.arrayOfLength 8 generate<byte>
-            |> Gen.map toFloat        
+            gen {
+                let! n = generate<DoNotSize<uint64>> |> Gen.map DoNotSize.Unwrap
+                return (float (n >>> 11)) * (1.0 / float (1UL <<< 53))
```
kurtschelfthout: I'm going to accept this as is, but looks like you don't need the `let!` here either (which will translate into `bind`) and can replace by another `Gen.map` call. This is also faster, also if you can get rid of the `gen {` computation expression altogether you'll save on another few thunks (like `Delay` etc).

Kharacternyk:
I didn't think about performance in general. I just reduced pressure on GC, because it's always good trick :). IMHO the fastest solution is implementing `Gen.choose64` in v3.0 and rewriting all of the `DoNotSize` gens to use it. Also if we had `Gen.choose64` we could generate 53 bits we need for a float mantissa in one `Gen.choose64` call instead of generating an uint64 and throwing away 11 bits or generating 26 and 27 bits in two calls of `Gen.choose`.

I've seen `Random.rangeInt64` on fscheck3 branch, so `Gen.choose64` shouldn't require any special efforts.

kurtschelfthout:
Using `map` instead of bind is a very similar suggestion, it will reduce GC pressure effectively. Generally in generators 'map' is faster than 'apply' is faster than 'bind'.

On fscheck3 branch I rewrote the random generator, I don't think the one on master generates the full int64 range (unlike the one on the fscheck3 branch).

Kharacternyk:
I know about the new random generator. My proposal is to implement Gen.choose64 in v3 release and reimplement 'DoNotSize' gens as well. I was surprised that fscheck3 has `rangeInt64` but does not have any API to call it.

There are many places in Arb.Default where bindings via gen comp. expr. may be replaced with Gen.map. Maybe, these will be gathered in PR soon.

kurtschelfthout:
> I know about the new random generator. My proposal is to implement Gen.choose64 in v3 release and reimplement DoNotSize gens as well. I was surprised that fscheck3 has rangeInt64 but does not have any API to call it.

Yeah, was trying to get the base stuff out first before changing the generators. But given the two branches have diverged so much already there is probably not much point in holding off anymore.

> There are many places in Arb.Default where bindings via gen comp. expr. may be replaced with Gen.map. 

Absolutely.

Issue #492, opened 2019-03-25
Standardize docs in Arb.Default


Kharacternyk:
Some current issues:
* Mention uniform distribution in sized integer (float, decimal) gens?
* Add info about shrinkers of number types?
* Mention that `Byte` gen is unrestricted by size?
* Mention probability of `null` (1/10 currently) in string gen?
* Mention that `Array` and other collections (`Set`, `Map`) length is determined by size?

Issues above are important because xml docs are the only source of information about Default arbs. 
Also, minor issue about style: *Generate* or *Generates*?

kurtschelfthout:
I think more clarity and info generally is definitely a Good Thing. So yes to all of the above!

> Also, minor issue about style: Generate or Generates?

I suggest "generates", because "generate" sounds like a command :)

As a template:

Generates `<type>` values `<size or other restrictions>`
`<other notes>`

For example
Generates uint16 values between 0 and size.


And I don't think we need to repeat "uniform distribution" every time, but just assume it unless we mention something else. Also lots of the comments seem to contain "arbitrary" like "Generates arbitrary uint16 values..." which I think we can also omit.

kurtschelfthout:
For info about shrinkers I'm less sure. Perhaps something like:

Shrinks values towards 0.

Or 

Shrinks values towards the empty list, and shrinks elements.

Suggestions welcome.

Kharacternyk:
I'm not sure how to properly describe integer shrinker. Maybe it is too complex to sumarize it in xml docs. It may be a paragraph on github.io docs. Collections as well.

About floats I think `Shrinker yields zero, absolute value of input and truncated input` will be fine.

Kharacternyk:
> `Shrinker yields zero, absolute value of input and truncated input`

Or use `origin` instead of `input`

Kharacternyk:
> And I don't think we need to repeat "uniform distribution" every time, but just assume it unless we mention something else.

Should it be written anywhere? Or it is already mentioned and I haven't noticed? To be honest, I thought output is concentrated on smaller values even when size is large when I used FsCheck first times.

kurtschelfthout:
> Should it be written anywhere? Or it is already mentioned and I haven't noticed? To be honest, I thought output is concentrated on smaller values even when size is large when I used FsCheck first times.

[...]

PR #494, opened 2019-03-28
Standardize docs in Arb.Default (#492).

Kharacternyk:


Kharacternyk:
Woah, looks like github thinks I wrote all of the TestData.fsx. What I've done wrong?

kurtschelfthout:
Yes, weird. It looks like it thinks you added it. Did you mark it for add somehow?

Kharacternyk:
>  Did you mark it for add somehow?

I'm not sure I understand. Is this file not tracked by git? I've added a paragraph to the docs but it seems I've done it a wrong way. If so how to do it properly?

kurtschelfthout:
I don't know how it happened. Perhaps putting the changes on a stash and then try to re-apply them?

Kharacternyk:
I did the exact things I had done before but via cmd instead of via Visual Studio and it works now. Closing it, #497 is a fixed version of it.

Kharacternyk:
Interesting that this is reproducable. I just can't do this right via VS. Looks like a bug.

Issue #444, opened 2018-07-14
Allow tuning of size distribution introduced in 2.10.0


glchapman:
From my experience with other QuickCheck libraries, I was surprised by the prevalence of ("", 0) in  this:
```
> Gen.sample 10 10 Arb.from<(string * int) list>.Generator;;
val it : (string * int) list list =
  [[("", 0); ("", 4)];
   [(null, 0); ("", 0); ("", 0); ("", 0); ("", 0); ("", 0); ("", 0); ("", 0);
    ("", 0); ("", 0)]; [("lb", 1)];
   [("", 0); ("", 0); (null, 0); ("", 0); ("", 4); ("", 0)];
   [("", 0); ("3", 1); ("", 0); (null, 0); ("", 0); ("", 0); ("", 0); ("", 0);
    ("", 0)]; [("", -1); ("", 0); ("", 0); ("", 0)]; [("", 0); ("", -1)];
   [(null, 0); ("", 0); ("", 0); ("", 0); ("", 0); ("", 0)];
   [("", 0); (null, 0); ("^", -1)]; [("QAz", 0)]]
```

I've briefly scanned the paper you reference in the release notes for 2.10.0, and, if I understand correctly, FSCheck is only using the fixedSize Distributor described in the paper. I think it would be useful to implement the other Distributors described, and provide some mechanism for selecting them when generating data.

kurtschelfthout:
I became convinced that the way the original quickcheck distributed size is less than ideal; what we're used to is not necessarily better.

Also, my sense is that adding distributors causes more complexity than it's worth. 

The options available in FsCheck today, and likely for the forseeable future, are:

- for coarse control, simply increase the size; 
- for fine-grained control over how size is distributed, write a custom generator using the size related functions on `Gen`.

kurtschelfthout:
Just to let you know I've changed my mind about what I said above. I now think we should somehow allow either the old or the new behavior for product-like types, and that it should be configurable. I need to think a bit of how to fit it into the API though, and it will go into v3.

witoldsz:
I am having similar issue, i.e.: having such a signature:
`rates: Map<ProviderId, RateObject list>` and `MaxTest=5` the FsCheck is generating, almost always, for 3 of 5 times the map with empty list.

On the other hand, if I change `rates` to:
`rates: RateObject list` and `MaxTest=5`
then FsCheck is generating 5 times a very long lists. Not even once the list is empty.

The `RateObject` looks like this: 
[...]

PR #495, opened 2019-03-29
Distribute the sizes of the generated elements in Gen.array2DOf.

Kharacternyk:
This PR makes the behaviors of array2DOf and arrayOf consistent.

Maybe there should be some wrapper types to allow users to generate lists and arrays the old way? (related: #444)

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -589,9 +589,12 @@ module Gen =
     /// Generates a 2D array of the given dimensions.
     //[category: Creating generators from generators]
     [<CompiledName("Array2DOf")>]
-    let array2DOfDim (rows: int,cols: int) (g: Gen<'a>)  = 
-        gen { let! arr1 = arrayOfLength (rows * cols) g
-              return Array2D.init rows cols (fun r c -> arr1.[cols*r + c]) }
+    let array2DOfDim (rows: int,cols: int) (g: Gen<'a>)  = sized (fun size ->
+        gen {
+            let length = rows * cols
+            let! sizes = piles length size
+            let! gens = Array.init length (fun i -> resize sizes.[i] g) |> sequenceToArr
+            return Array2D.init rows cols (fun r c -> gens.[cols*r + c]) })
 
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Hm, yes I hadn't thought of using wrapper types. Or maybe just some extra Arbitrary.Default methods would be fine too.

kurtschelfthout:
Anyway I'm not convinced we should do this for 2D arrays. 2D arrays do not usually contain nested  structures.

Kharacternyk:
There are two ways of representing a multi-dimensional array in .NET: nested arrays and "real" 2,3...d arrays. It is an implementation detail, logically they are the same (I mean they equally represent matrixes of fixed width and height. Nested arrays can be used for some other things though).

IMHO FsCheck should distribute a size the same way in both of these cases.

kurtschelfthout:
I don't agree it's an implementation detail. 2D arrays are mostly used to represent matrices, and mostly contain primitive types (floats etc). I'd never represent a matrix as a nested array. 

The distribution of the size is actually a pretty complicated question. I'm not sure absolute consistency is the answer here. Size is most importantly a mechanism to control the time it takes to generate complex values. For recursive values size bounds the number of recursions. For primitive types (and also complex types) on the other hand it's just a heuristic to make "smaller" test failures happen "sooner". But the latter becomes a lot less important when shrinking is in play. 

The trade-off here is that smaller values arguably means less test coverage.

The 1D size distribution mechanism was changed from giving each element of a list n-1 size to making the sum of the sizes add up to n, because 1D data structures frequently contain complex values, and more importantly, can contain nested 1D data structures (e.g. `list<array<T>>`). In those cases it becomes important to distribute the size in such a way that individual elements get a much lower size than size-1.

But my sense is that for 2D arrays it's much less the case that they contain complex elements, and I can't remember ever seeing a thing like `array[,]<list<T>>` or something like that. So that would indicate that for 2D arrays we're fine to assign size-1 to each element of the 2D array, and thereby increase test coverage somewhat.

Does that make sense?

Kharacternyk:
I understand your point. Yeah, it makes sense. I'm going to add the shortened version of all of the above to xml docs in #494.

PR #496, opened 2019-04-02
Add Gen.scale to avoid some boilerplate code.

Kharacternyk:
Inspired by the Scalacheck one.

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -612,6 +612,9 @@ module Gen =
     [<CompiledName("Apply")>]
     let apply (f:Gen<'a -> 'b>) (gn:Gen<'a>) : Gen<'b> = apply f gn
 
+    // Modify a size using the given function and invoke the given Gen with this size
+    let scale f g = sized (fun size -> resize (f size) g)
+    
     ///Promote the given function f to a function generator. Only used for generating arbitrary functions.
     let internal promote f = Gen (fun n r -> fun a -> let (Gen m) = f a in m n r)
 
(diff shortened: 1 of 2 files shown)


Kharacternyk:
An example of usage: if I want to generate the couples of integers to represent a height and a width of some type of a game board, the count of squares on board will increase as size^2 by default. To avoid this, the following code might be used:
```f#
let board1 =
    (fun n -> Gen.resize (n |> float |> sqrt |> int) Arb.generate<int>)
    |> Gen.sized 
    |> Gen.two
```
and now it looks like this:
```f#
let board2 =
    Gen.scale (float >> sqrt >> int) Arb.generate<int>
    |> Gen.two
```

Kharacternyk:
If the idea isn't rejected, I'll write the xml docs, some more tests and maybe some "overloads" for list, etc.

kurtschelfthout:
I think that's a good idea, maybe call it `scaleSize` so it's clear what is being scaled.

> maybe some "overloads" for list, 

I don't understand that, why would you need additional overloads?

Kharacternyk:
> I don't understand that, why would you need additional overloads?

I mean something like `Gen.scaleList` which require a list of sizes as an argument. Also we may provide `Gen.scaleSize2` of signature `(int -> int * int -> int) -> (Gen * Gen) -> Gen<a, a>`

kurtschelfthout:
> It's a bit controversial, though. New API is good but defining a function that would be used only few times by users is redundant. That's why I haven't added all of the above yet. Thoughts?

Yeah I agree - don't think any more overloads pass the bar.

kurtschelfthout:
This looks ready to merge then?

Kharacternyk:
Yeah, I think so.

Issue #501, opened 2019-08-02
Support for struct C# record


kmusick:
The following class is generated without issue:

``` fsharp
type ClassRecord(a: int) =
    member this.A = a
    
Arb.from<ClassRecord>
|> Arb.toGen
|> Gen.sample 50 10
```

If you change this to a struct, generation fails:

``` fsharp
[<Struct>]
type StructRecord(a: int) =
    member this.A = a

Arb.from<StructRecord>
|> Arb.toGen
|> Gen.sample 50 10
```

> The type Main+StructRecord is not handled automatically by FsCheck. Consider using another type or writing and registering a generator for it.

This would be useful for struct records in C#:

``` csharp
public struct StructRecord
{
    public int Id { get; }
    public StructRecord(int id)
    {
        Id = id;
    }
}
```

Kharacternyk:
When I was working on the enum reflective generator and shrinker I noticed this function:
https://github.com/fscheck/FsCheck/blob/d06b0a70a7159e979324e375b6f1c53b2f7c38a7/src/FsCheck/Reflect.fs#L21-L27
I was wondering then what is the purpose of `typeinfo.IsClass` check. Maybe just removing it solves the issue? Or am I missing something?

kurtschelfthout:
Yep, this is a very reasonable request. 

I don’t think the IsClass serves any particular technical purpose. The original contributor of this code was just conservative in what was supported and tested for expedience. Want to give it a try?

Kharacternyk:
Turns out that removing the `IsClass` check doesn't solve the problem. The current algorithm doesn't handle structs and throws an exception:
```
System.ArgumentException : Expression of type 'FsCheck.Test.Arbitrary+FakeValueRecord' cannot be used for return type 'System.Object'

StackTrace:	
at System.Linq.Expressions.Expression.ValidateLambdaArgs(Type delegateType, Expression& body, ReadOnlyCollection`1 parameters)
   at System.Linq.Expressions.Expression.Lambda[TDelegate](Expression body, String name, Boolean tailCall, IEnumerable`1 parameters)
   at FsCheck.Reflect.getCSharpRecordConstructor(Type t) …
```
This issue requires some refactoring.

PR #502, opened 2019-08-03
Added struct record generation support

kmusick:
#501 
Adds support and tests for struct record generation


kurtschelfthout:
Thanks!

PR #505, opened 2019-08-12
FEAT: add C# Prop.When

stijnmoreels:
### More C#-friendly conditional property that holds evaluation
Before, we had to create a separate degate to specify properties that should only evalute when the condition holds. Example:

```csharp
Prop.ForAll<int>(a => new Func<bool>(() => 1 / a == 1 / a).When(a != 0))
```

With this PR we can specify it a bit more dev-friendly:
```csharp
 Prop.ForAll<int>(a => Prop.When(a != 0, () => 1 / a == 1 / a))
```

### DIFF
--- a/examples/FsCheck.CSharpExamples/Program.cs
+++ b/examples/FsCheck.CSharpExamples/Program.cs
@@ -66,8 +66,11 @@ namespace FsCheck.CSharpExamples
             Prop.ForAll<int, int[]>((x, xs) => xs.Insert(x).IsOrdered().When(xs.IsOrdered()))
                 .QuickCheck("Insert");
 
-            Prop.ForAll<int>(a => new Func<bool>(() => 1 / a == 1 / a).When(a != 0))
+            Prop.ForAll<int>(a => Prop.When(a != 0, () => 1 / a == 1 / a))
                 .QuickCheck("DivByZero");
+
+            Prop.ForAll<short[]>(xs => Prop.When(condition: xs.Length > 0, assertion: () => xs.Length > 0))
+                .QuickCheck("When non-empty array ==> Assert non-empty array");
 
             //counting trivial cases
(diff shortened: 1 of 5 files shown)


kurtschelfthout:
Thanks! am on holiday at the moment. Will take a look in a few weeks.

kurtschelfthout:
Yes! Really sorry, dropped the ball on this. Will take a look tonight.

kurtschelfthout:
I like the `Prop.When` and would like to merge it.

Not sure I like the NOT-operator `!!`, did you actually have a "real-world" use for it?

jzabroski:
> Not sure I like the NOT-operator `!!`, did you actually have a "real-world" use for it?

It doesn't seem useful with a default Arb.Int32.  It would be useful if you could customize the Arb.

kurtschelfthout:
@jzabroski can you elaborate? It inverses the outcome of the property (i.e. if it returns false, the test passes), I don't see what customizing the `Arb` has to do with it...

jzabroski:
@kurtschelfthout  I agree, my comment doesn't make sense.  I think I was reading/replying too fast.

@stijnmoreels The negation of generalization isn't to negate "for all", but rather to prove "there exists".

The negation of “For all _x_, _S(x)_” is logically equivalent to
“There exists an _x_ such that _not_[_S(x)_].”

The reason is that it is not given that the inverse of a proposition is true.  Calling this an inverse is not correct.

kurtschelfthout:
> The negation of generalization isn't to negate "for all", but rather to prove "there exists".

That's a good point. 

@stijnmoreels maybe you could factor out the two things being proposed in separate PRs?

kurtschelfthout:
Or perhaps just factor out the `Prop.When` as I think that's not contentious, while we discuss the negation here.

PR #511, opened 2020-02-27
Support setting `null` to the Replay property

eiriktsarpalis:
Explicitly setting the `PropertyAttribute.Replay` property to `null` is currently broken. This is useful in scenaria where you are using a global constant to define the seed for a suite of property tests.


kurtschelfthout:
Thanks!

Issue #421, opened 2018-01-09
Harmonize Config/Configuration/PropertyAttribute


kurtschelfthout:
As noted in #420 we have about a handful of configuration types that have slightly different property names.

As much as possible we should limit the number of different configuration types. At least make all the names the same to avoid confusion.

Also, the basic `Config` type https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Runner.fs#L39 should become a POCO mutable type - that means we can probably get rid of a few wrappers. And also, F# record types in a public API are basically a big no-no, as they can't be changed without breaking compatibility.

This should happen on the fscheck3 branch, and where possible add/deprecate in v2.

ploeh:
Why do you want to make it mutable?

kurtschelfthout:
> Why do you want to make it mutable?

So we can have just one `Configuration` type that is conveniently usable from both F# and C#.

ploeh:
It doesn't have to be mutable in order to be usable from C#...

kurtschelfthout:
_conveniently_ usable ;) from both langs and easily extensible. I guess you're referring to using optional parameters but that gets unwieldy pretty quickly I think.

ploeh:
No, I wasn't thinking of optional parameters. What you can do in C#, and how I'd design C# APIs even in the absence of F#, would be something like a Fluent Builder API that would enable people to write something like:

```csharp
Configuration.Default.WithMaxTest(1000);
```

The problem with using F# records for public APIs has nothing to do with immutability, but that you can't add constructor overloads with default values. One can address that, however, by not exporting the internal structure of the record:

```fsharp
type Config = private Config of (maxTest : int, maxFail : int (* ... *))
```

(It's possible that I got the details of the syntax wrong; I haven't tried to compile that snippet...

kurtschelfthout:
Yes, hadn't considered that - fluent builder will work. 

The problem I have with record types is that they can't be changed without breaking backwards compatibility, specifically breaking the constructor. There have been various proposals to F# to address that, but they've been consistently rejected.

Afaik you can't make a public record type with a private constructor. 

They are also annoying to use from C#. 

Going forward I want all record types removed from the public API and am banning record types from appearing in the public API.

We can still use record types for the internals if useful.

ploeh:
> Afaik you can't make a public record type with a private constructor.

IIRC, you're right. One can, however, hide them in single-case DUs.

LaurenceJKing:
[...]

PR #517, opened 2020-04-27
Harmonize config

LaurenceJKing:
#421


kurtschelfthout:
Looks like global.json needs a .NET core version bump for AppVeyor:

A compatible installed .NET Core SDK for global.json version [3.1.101] from [C:\projects\fscheck\global.json] was not found
Install the [3.1.101] .NET Core SDK or update [C:\projects\fscheck\global.json] with an installed .NET Core SDK:
  1.1.14 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.805 [C:\Program Files\dotnet\sdk]
  2.2.402 [C:\Program Files\dotnet\sdk]
  3.0.103 [C:\Program Files\dotnet\sdk]
  3.1.201 [C:\Program Files\dotnet\sdk]


3.1.201 is the one to go for here.

src/FsCheck/Runner.fs
```diff
@@ -48,34 +48,74 @@ type ParallelRunConfig =
...
+                ///What to print every time a counter-example is successfully shrunk
+                EveryShrink   : list<obj> -> string 
+                ///The Arbitrary instances on this class will be merged in back to front order, i.e. instances for the same generated type at the front
+                ///of the list will override those at the back. The instances on Arb.Default are always known, and are at the back (so they can always be
+                ///overridden)
+                Arbitrary     : list<Type>
+                ///A custom test runner, e.g. to integrate with a test framework like xUnit or NUnit. 
+                Runner        : IRunner
+                ///If set, inputs for property generation and property evaluation will be run in parallel. 
+                ParallelRunConfig : ParallelRunConfig option
```
kurtschelfthout, line 77: Can you move the docs to the `WithX` methods?

src/FsCheck/Runner.fs
```diff
@@ -48,34 +48,74 @@ type ParallelRunConfig =
...
+                |} with
+                ///Returns the values for the current Config
+                member this.Values =
+                    let values (Config config) = config
+                    this |> values
```
kurtschelfthout, line 82: What's the return type of `Values` - is it effectively Tuple?

kurtschelfthout:
Overall this is looking really good! Thanks for your efforts.

src/FsCheck/Runner.fs
LaurenceJKing: It's an anonymous record. [Anonymous Records](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/anonymous-records).

I was torn between exposing the record, or adding value members:
`member private this.Values =
    let values (Config config) = config
    this |> values
member this.MaxTest = this.Values.MaxTest`

kurtschelfthout: Yeah, I'm just not sure if it generates a proper .NET object with properties, or whether it generates a `Tuple` with some metadata for the compiler to hook into. My main concern is that, historically for a record type, adding a new configuration parameter has been a breaking change in the SemVer sense, because effectively that adds a parameter to the ctor of the underlying .NET class F# record types are compiled to. So I'm keen to know whether the actual type (meaning, whatever an anonymous record compiles to) that is being returned here can have new properties added without being a breaking change.

If you don't know this is something that can be dealt with later, but just thought I'd bring it to your attention.

LaurenceJKing: It looks like it compiles to a .Net class rather than a tuple. I have updated this so that the whole record is private and the values are exposed via members. Hopefully that solves the issue?

src/FsCheck/Runner.fs
LaurenceJKing: I've added comments to the WithX methods and moved the comments from the record.

I also noticed that the comments were not being generated for the anonymous record, so I copied the fields (and their comments) into members of Config.

kurtschelfthout:
Thank you very much for this!

PR #518, opened 2020-06-03
Added Comments to better explain C# Tests

SpocWeb:
Hi while trying to understand your C# Examples I added some Comments to better explain them. I think they would be helpful for the next reader. 
Personally I would have loved to see such brief explanations, especially in sample code.


SpocWeb:
AppVeyor build fails, because only .NET 3.1.202  is available not 3.1.101 as specified in global.json. Should I change that?

kurtschelfthout:
This happens too regularly :) yes feel free to update the global.json

Consider adding     `"rollForward": "latestMajor"` as described in  https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x so we don't run into this again. (I can do it in a separate commit too, but while you're there. Up to you.)

Thanks for the comments!

SpocWeb:
OK I upgraded the Version and added the "rollForward": "latestMajor" Section, let's see whether it builds.

kurtschelfthout:
Ok, travis has an issue but I'll sort that out on my end.

Thanks!

SpocWeb:
travis-ci doesn't seem to run. Can you please look into that?

Issue #514, opened 2020-03-18
Using FsCheck.Xunit in C#, how do I clean up/tear down after a test has been executed?


CAP-3Shape:
I am using FsCheck - specifically the FsCheck.Xunit package - in my C# code. I want to run some setup before my tests execute and some cleanup/teardown code after my tests have all run.

NUnit has a [CleanUp] attribute; Xunit uses _Dispose_ to clean up. In some frameworks I think it's called [TearDown]. Is there anything of this sort in FsCheck.Xunit?

When I rewrote one Xunit test class to use FsCheck, I could see that my _Dispose_ method no longer got executed.

In the [documentation ](https://fscheck.github.io/FsCheck/reference/index.html) there is something called "TearDown", but it is completely opaque to me how it is supposed to be used.

I [posted this on Stack Overflow](https://stackoverflow.com/questions/60702341/using-fscheck-xunit-in-c-how-do-i-clean-up-tear-down-after-all-tests-have-been) and Kurt Schelfthout asked me to report it as an issue here. 

Is it a bug in FsCheck.Xunit? Ought it to execute my _Dispose_ method like a regular Xunit test?

I am using: 

- FsCheck.Xunit 2.14.1
- .NET core 3.1
- C# 8.0
- Visual Studio 2019 version 16.4.6

kurtschelfthout:
Yes, it looks like after replacing FactAttribute with Property, xunit no longer calls Dispose. 

I’ll have to take a look why. 

Meanwhile you may be able to work around by not using FsCheck.Xunit at all, by defining properties inside a normal Xunit Fact method, using eg QuickCheckThrowOnFailure and the Prop API.

ploeh:
Wouldn't that cause the property to run (by default) 100 times, and then do a single cleanup?

CAP-3Shape:
> Wouldn't that cause the property to run (by default) 100 times, and then do a single cleanup?

It would be great to have a way to clean up after each run AND a way to clean up after all runs, similar to the [TestCleanup] and [ClassCleanup] that some test frameworks have. I don't know how feasible that is in Xunit.

ploeh:
IIRC xUnit.net has nothing like `[ClassCleanup]`...

kurtschelfthout:
> Wouldn't that cause the property to run (by default) 100 times, and then do a single cleanup?
Indeed. I thought that was the original question.

[...]

PR #519, opened 2020-06-05
Call Dispose after Xunit test where test class inherits IDisposable

LaurenceJKing:
#514 

This pull request aims to resolve the bug uncovered in issue 514 (Dispose not called from IDisposable class when using xunit).

### DIFF
--- a/src/FsCheck.Xunit/PropertyAttribute.fs
+++ b/src/FsCheck.Xunit/PropertyAttribute.fs
@@ -184,8 +184,16 @@ type PropertyTestCase(diagnosticMessageSink:IMessageSink, defaultMethodDisplay:T
         let summary = new RunSummary(Total = 1);
         let outputHelper = new TestOutputHelper()
         outputHelper.Initialize(messageBus, test)
-        let testExec() =
-            
+
+        let dispose testClass =
+            match testClass with
+            | None -> ()
+            | Some obj ->
+                match box obj with
+                | :? IDisposable as d -> d.Dispose()
+                | _ -> ()
+
+        let testExec() =            
             let config = this.Init(outputHelper)
             let timer = ExecutionTimer()
             let result =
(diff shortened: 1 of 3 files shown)


ploeh:
As in #514, I'm still concerned that this only disposes _once_, while each property runs (by default) 100 times.

What problem does this address?

I'm not dead set against this change, but I think that maintainers (which is mostly @kurtschelfthout, so my voice shouldn't count for much) should be conservative when it comes to adding behaviour that may not be well understood. There's always a risk of painting oneself into a corner... I can't tell if there's a risk of that here.

src/FsCheck.Xunit/PropertyAttribute.fs
```diff
@@ -205,6 +213,7 @@ type PropertyTestCase(diagnosticMessageSink:IMessageSink, defaultMethodDisplay:T
 
                     match xunitRunner.Result with
                           | TestResult.True _ ->
+                            dispose target
```
ploeh, line 216: Do understand this correctly, that the code only disposes the `target` if the result is `True`?

Isn't the desired behaviour that the object should always be disposed, regardless of outcome?

LaurenceJKing: My mistake. I thought xunit only called dispose after a successful test, but it turns out I was wrong. I'll amend amend this to always dispose

ploeh:
Apart from my concerns, _if_ we want this behaviour, I think that the pull request looks fine.

kurtschelfthout:
@ploeh Thanks very much for your considered comments. As I see it, this PR addresses a _bug_ with how `PropertyAttribute` is wired into xUnit - I consider it a bug because if you have a disposable class with a `FactAttribute` method in it, `Dispose` will be called (once). Replace the `FactAttribute` with `PropertyAttribute` and now suddenly `Dispose` is no longer called. That is counter-intuitive. 

The ability to clean up after every FsCheck test is an interesting feature request, but is I believe orthogonal, and I wouldn't want FsCheck to call `Dispose` after every test. That would be similar to doing for a plain old object:

```fsharp
let o = new MyObject()
for i in [1..100] do
    o.MyMethodCall()
    o.Dispose()
```
which seems weird to me - `Dispose` is to `new` like `free` is to `malloc` in less GC-englightened languages ;)

I hope that makes sense.

PR #531, opened 2020-09-06
Less verbose C# examples

kant2002:
I found that C# examples is more scary to look at. It some places written in modern C# features it would be still understandable for the audience and at the same time, less boilerplate


kurtschelfthout:
Nice, thanks!

Issue #549, opened 2021-02-17
Unable to access constructor of DiscardException with optimized build.


OkkeHendriks:
When running the following test in a visual studio project which is unoptimized the test raises the expected `DiscardException`.
```F#
[<Xunit.Fact>]
let testUnoptimized () =
    FsCheck.Prop.discard ()
    true
```
When running the exact same test in a project which **is** optimized, the following unexpected exception is raised:
```
  Message: 
    System.MethodAccessException : Attempt by method 'test.testOptimized()' to access method 'FsCheck.Testable+DiscardException..ctor()' failed.
  Stack Trace: 
    test.testOptimized() line 6
```
![image](https://user-images.githubusercontent.com/3514136/108270371-a6e0f000-716f-11eb-89b8-82f88168ce9f.png)

This is **not** what happens when I run the `FsCheck.Test` project tests. There, Prop.discard () works as expected and as far as I can tell this is also with optimization on.

The above example, in both a project which has optimization enabled and in one with optimization disabled, is available here:
https://github.com/OkkeHendriks/fscheck_discard_issue

Does someone see my stupid mistake or is more going on?

_Version info_:
I tried to use the same .NET framework version (4.5.2) and other assemblies/packages as in the `FsCheck.Test` project on master.
```
$ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.102
 Commit:    71365b4d42

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.102\

Host (useful for support):
  Version: 5.0.2
  Commit:  cb5f173b96

.NET SDKs installed:
  2.1.811 [C:\Program Files\dotnet\sdk]
  3.1.101 [C:\Program Files\dotnet\sdk]
  5.0.102 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.23 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #550, opened 2021-02-18
Prevent inlining of Prop.discard

OkkeHendriks:
This is done as a workaround, as discussed at https://github.com/fscheck/FsCheck/issues/549, in order to prevent a MethodAccessException of the internal `DiscardException` constructor, which is caused by the compiler too aggressively optimizing the call.

### DIFF
--- a/src/FsCheck/Prop.fs
+++ b/src/FsCheck/Prop.fs
@@ -6,6 +6,7 @@ module Prop =
     open Testable
     open System
     open System.ComponentModel
+    open System.Runtime.CompilerServices
 
     ///Quantified property combinator. Provide a custom test data generator to a property.
     [<CompiledName("ForAll")>]
@@ -138,6 +139,7 @@ module Prop =
         property testable
 
     [<CompiledName("Discard")>]
+    [<MethodImpl(MethodImplOptions.NoInlining)>]
     let discard() = raise DiscardException
 
 ///Operators for Prop.
--- a/src/FsCheck/Prop.fs
+++ b/src/FsCheck/Prop.fs
@@ -139,6 +139,8 @@ module Prop =
         property testable
 
     [<CompiledName("Discard")>]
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
Thanks very much for the prompt PR!

Would you mind adding either a comment in the code explaining the issue, and/or a link to this PR to that attribute? These kinds of things tend to be easily forgotten and then regress...

OkkeHendriks:
I added the comment, however, while typing it I checked whether the `DiscardException` is actually internal but it is inside the private Testable module: https://github.com/fscheck/FsCheck/blob/02c5ad6be05eac65e536ac86ec0086e98d3cb3a8/src/FsCheck/Testable.fs#L101  

So I do not yet understand why this works in `FsCheck.Test`, but this might have something todo with me not completely understanding the accessor modifiers of modules i.c.w. `InternalsVisibleTo`.

I did notice there is also an internal Prop module inside the Testable module, might have something todo with it? (the discard function is in the public Prop module): https://github.com/fscheck/FsCheck/blob/02c5ad6be05eac65e536ac86ec0086e98d3cb3a8/src/FsCheck/Testable.fs#L120

kurtschelfthout:
> So I do not yet understand why this works in FsCheck.Test, but this might have something todo with me not completely understanding the accessor modifiers of modules i.c.w. InternalsVisibleTo.

Yes, it's confusing - I believe `private` modules in F# actually get compiled to `internal` classes, and so `InternalsVisibleTo` works.

kurtschelfthout:
Thanks very much for the PR!

OkkeHendriks:
No problem!

I don't know what the release strategy/schedule is but do you expect to make a release soon?
Then I can decide if we wait or temporarily have to work around it.

kurtschelfthout:
Can do it right now, let me just update the release notes.

PR #559, opened 2021-04-06
Allow xUnit Properties attribute to be used at assembly level

LaurenceJKing:
I hope to add similar functionality to the NUnit project too, but I'm less familiar with NUnit, so this will take a bit more investigation.


kurtschelfthout:
Very nice, thank you!

Since you mention it I'll wait a bit for the NUnit additions as well to push a release? But feel free to say otherwise, it's not a problem at all.

LaurenceJKing:
The NUnit work is taking a bit longer than I'd hoped, so feel free to wait or push at your discretion.

Issue #525, opened 2020-07-29
Consider removing the restriction that the generator is public.


teo-tsirpanis:
Consider a library like this:

```
[<assembly:InternalsVisibleTo("MyLibraryTests")>]

type internal MyDataStructure = ...

type LibraryAPI = ...
```

This library contains an internal data structure class, which despite being an implementation detail (as opined in the fourth bullet of https://github.com/fscheck/FsCheck/issues/152#issuecomment-141634952), but it's still a type whose behavior is worth verifying by itself, but irrelevant to the purpose of the library to warrant making it public.

Now, let's write a generator:

```
// We are at the MyLibraryTests assembly.

type Generators() =
    static member MyDataStructure() = Arb.generate |> ...
```

The code above raises a compiler error because `Generators` is public but one of its methods returns an internal type. We could make `Generators` internal but FsCheck would complain, forcing us to make `MyDataStructure` internal.

---

The best way to actually solve such problems is to make FsCheck allow nonpublic generator classes. It might be done with [a single-line change in `Runner.fs`](https://github.com/fscheck/FsCheck/blob/master/src/FsCheck/Runner.fs#L284) but I'm not sure whether it would suffice.

kurtschelfthout:
See https://github.com/fscheck/FsCheck/issues/401

kurtschelfthout:
Closed as dup.

Issue #436, opened 2018-05-04
Exception when comparing generated map containing private DU


jbezak:
I ran into an exception when running following test:

> Message: System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
---- System.ArgumentException : The type 'SiteAgent.MessageGateway.Server.Tests.Foo+Bar' is an F# union type but its representation is private. You must specify BindingFlags.NonPublic to access private type representations.
Parameter name: unionType

```
open FsCheck
open FsCheck.Xunit

module Foo =
    type Bar = private Bar of string
    let create bar = Bar bar

module DataTests =
    [<Property>]
    let ``empty = generated nonEmpty`` (bars: Map<Foo.Bar, string>) =
        not (bars.IsEmpty) ==> lazy
        let emptyMap : Map<Foo.Bar, string> = Map.empty
        emptyMap = bars
```
When the map is created inside the test it fails correctly without the exception:
```
    [<Property>]
    let ``empty = nonEmpty``() =
        let bar = Foo.create "bar"
        let nonEmptyMap = Map [ bar, "bar" ]
        let emptyMap : Map<Foo.Bar, string> = Map.empty
        emptyMap = nonEmptyMap
```

> map [] = map [(Bar "bar", "bar")]
false
Expected: True
Actual:   False

ploeh:
That seems like reasonable behaviour to me. Since you're reporting this, I suppose your expectation was different. What do you think should happen instead?

ploeh:
That seems like reasonable behaviour to me. Since you're reporting this, I suppose your expectation was different. What do you think should happen instead?

jbezak:
I expected that the ``empty = generated non-empty`` test would produce a failing test, not a test throwing an exception, same as the non-generated example.

kurtschelfthout:
FsCheck does not generate private union/record types.  See #186.

The idea is that those are private for a reason, and so it can't just generate arbitrary data, there are some preconditions that need to be satisfied.

jbezak:
If I understand the pull request you've mentioned correctly it actually adds the functionality to generate private DU. That is also what I am seeing when debugging the tests.

I realized that the exception does not occur during the generation but during shrinking for failed tests.
This test is passing but when it fails, shrinking will not help me to find the failing values.
  ```
    [<Property>]
    let ``generating nonEmpty`` (bars: Map<Foo.Bar, string>) =
        not (bars.IsEmpty) ==> lazy
[...]

Issue #401, opened 2017-10-04
Problems when generating internal types


cmeeren:
I want to test `internal` stuff (using `InternalsVisibleTo` on the main assembly). Specifically, I want FsCheck to generate instances of internal types. This seems to be a bit hit-and-miss, and mostly seems to fail. For example, generating values of an internal union gives `SystemArgumentException` with message *X is an F# union type but its representation is private. You must specify BindingFlags.NonPublic to access private type representations.* Generation of some other internal types seem to loop until the test times out and gets aborted.

Is this a known issue? Is it deliberate? Would it be possible to make FsCheck able to generate internal types?

Currently I'm just leaning towards making everything public, but that means the public API of the main assembly is rather more noisy than it needs to be.

cmeeren:
Nevermind the infinite loop part; it seems that it was caused by FsCheck not picking up `internal` Arbitrary-returning members on the class I specified in the `PropertiesAttribute`, and thus not using my designated generators for a few recursive types.

kurtschelfthout:
I'm on the fence about getting FsCheck to take internal and private constructors into account. I usually use that to signal that the ctor doesn't do validation or doesn't necessarily uphold an invariant. I assume that the same problem would exist when FsCheck would be trying to call those ctors. So it can cause some confusion - "why isn't FsCheck only using my public ctor, it's generating bogus instances by breaking my encapsulation" type issues.

cmeeren:
In that case surely FsCheck could always use a public ctor if both public and internal ctors exists?

kurtschelfthout:
> In that case surely FsCheck could always use a public ctor if both public and internal ctors exist?

That doesn't seem very transparent either, the generation strategy would change based on access modifiers... 

I would consider a configuration option or some such. Unfortunately currently the `Config` type is a record type, which basically  means that every addition to it is a breaking API change. Small mistake, high price 😒 

In the shorter term, it would be possible to add an overload of `Arb.Default.Derive` that takes an option to also use internal/private constructors. You'd then still need to define the arbitrary instances, but they would be a one-liner each.

It probably won't be high in my priority list. I would accept a PR that adds it provided the current behaviour remains the default and the configuration for the new behavior is explicit. It shouldn't be hard to do, some plumbing to get the reflection calls configured correctly.

ploeh:
[...]

PR #567, opened 2021-05-04
Allow shrinking private unions

svdijk:
This PR allows FsCheck to shrink values of unions with private constructors, making the behavior for unions with private constructors consistent with that for records with private constructors.

FsCheck could already derive generators for records and unions with private constructors (for unions, this functionality was added in PR #186). It could also shrink values of records with a private constructor, but not values of unions with a private constructor. Several people have run into that, see issues #401, #436 and #525.

Whether automatically deriving generators and shrinkers for types with a private constructor should be done in general is something that could be discussed (this is also touched upon in issue #436), but since FsCheck already does generator derivation for such types and also does shrinker derivation for such record types, I believe being consistent for such union types is important enough to warrant this PR regardless of said discussion.


kurtschelfthout:
Thank you!

PR #576, opened 2021-07-25
Support generating C# records

### LINKED ISSUE #564: Reflection for records fails with C# 9.0
In C# 9.0 compiler generates properties of `record` classes as init-only. It appears that `isCSharpRecordType` does not accommodate for this and fails to recognize such classes as records. To be more specific, reflection-based generation fails for this:

```
public record RgbColor(byte Red, byte Green, byte Blue) {}
```

This works:

```
public record RgbColor(byte Red, byte Green, byte Blue)
{
    public byte Red { get; } = Red;

    public byte Green { get; } = Green;

    public byte Blue { get; } = Blue;
}
```

This post suggests the way to detect init-only properties through reflection: https://alistairevans.co.uk/2020/11/01/detecting-init-only-properties-with-reflection-in-c-9/. I believe that `isCSharpRecordType` must be fixed.

kurtschelfthout:
Closes #564 

The actual change is just to `isCSharpRecordType`, but had to update and adjust test projects to use net 5.0 to be able to conveniently use the C# record type feature in a C# test project. This caused quite a bit of downstream build shenanigans.


paoloambrosio:
This is great! Tried this version and `Arb.Default.Derive` works as expected (see #578).

kurtschelfthout:
Travis apparently has some issues connecting to this repository now that travis-ci.org no longer works. Going to ignore.

PR #588, opened 2021-08-30
Add FsCheck.Xunit.TestOutputRunner.

### LINKED ISSUE #509: Before submitting a PR: XUnit test runner that can be used directly
Hi,

As [discussed on Twitter](https://twitter.com/breki74/status/1203632900069044224), I needed an XUnit runner that actually writes all of its output using XUnit's `ITestOutputHelper` and that can be used directly, not just through `Property` attributes, so I implemented one by (mostly) cannibalizing existing code from FsCheck: https://github.com/breki/demeton/blob/master/Demeton.Tests/BetterFsCheckRunner.fs

Here's a sample on how it is used: https://github.com/breki/demeton/blob/master/Demeton.Tests/PropertiesHelp.fs#L46

While I could add the whole code as a PR, I think it would be a better idea to discuss how best to integrate it - do I just merge the code with the existing XUnit runner or implement it separately?

kurtschelfthout:
Closes #509 

@breki what do you know, I finally got to this. From your example:
- I removed the `verbose` flag because that ought to be controlled via the `Config` by setting `Every` and `EveryShrink` and so I felt it was redundant
- Made it into a proper type so it's easier to access for C# users, and also you can now inherit it and call the relevant `super` methods to do the default thing if you want (like e.g. bringing `verbose` back....)

Let me know if you have any thoughts.

### DIFF
--- a/src/FsCheck.Xunit/FsCheck.Xunit.fsproj
+++ b/src/FsCheck.Xunit/FsCheck.Xunit.fsproj
@@ -8,6 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="AssemblyInfo.fs" />
+    <Compile Include="Runner.fs" />
     <Compile Include="PropertyAttribute.fs" />
     <Compile Include="CheckExtensions.fs" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


breki:
That is great news! I haven't done much work in F# lately, but I'm planning to return to one of my pet project that will use FsCheck. I will let you know how it goes.

Issue #612, opened 2022-06-11
Version 3.x, register Arbitraries directly with ArbMap


farlee2121:
## Feature
I love the move to ArbMap in v3, over the thread-level registry.

It'd be great if we could directly register Arbitrary instances with ArbMap instead of needing an intermediary type with static members.

For example
```fsharp
let customArb = //...
ArbMap.defaults |> ArbMap.with customArb
```
This would allow users to simply define and use an Arbitrary all in one local scope.
This seems especially relevant now that ArbMap enables users to override type generation in a limited scope. It seems likely users would define an arbitrary only meant to be used in constructing some parent arbitrary.

## Implementation Ideas
It seems TypeClass currently bears the responsibility for tracking the map of types to Arbitrary instances, and it tracks them as `Map<InstanceKind,MethodInfo>`.

In order to directly register Arbitraries, this map would need to change.

I haven't quite grasped TypeClass yet, but here's my idea.
- It makes sense to me that ArbMap is responsible for tracking instances of arbitraries, maybe `Map<InstanceKind,IArbitrary>`. 
- TypeClass can be responsible for knowing what a match looks like / the light construction it seems to do
  - maybe `arbMap |> TypeClass.getInstance type`
- Discovery then also probably moves somewhere else, and it outputs an ArbMap 
  - maybe `let arbMap = Discover.factoriesOnType<type>`  (or maybe it outputs a `Map<Type, IArbitrary>`) 
  - If we don't want to invoke the factories right away, then we could create some internal lazy Arbitrary derivative that invokes the matched static factory on first call

farlee2121:
I've investigated further and was able to create a pretty simple proof of concept.

[Branch comparison](https://github.com/fscheck/FsCheck/compare/fscheck3...farlee2121:direct-arb-registration?expand=1)

The bulk of it is this one method added to TypeClass
```fsharp
member this.MergeMethod(factory: System.Func<'TypeClass>) = 
    let typeClass = this.Class
    let toRegistryPair (m:MethodInfo) =
        match m.ReturnType with
        | GenericTypeDef typeClass args when args.Length <> 1 -> 
            failwithf "Typeclasses must have exactly one generic parameter. Typeclass %A has %i" typeClass args.Length
        | GenericTypeDef typeClass args ->
            let instance = args.[0]
            (InstanceKind.FromType instance,m)
        | _ -> failwith "Lambda did not return a compatible type for this type class"
            
[...]

PR #613, opened 2022-06-30
Direct arb registration

farlee2121:
#612


kurtschelfthout:
Thanks a lot - I will try to take a look this weekend, but might have to wait till next week.

farlee2121:
No rush. Thanks for working through it with me!

src/FsCheck/ArbMap.fs
```diff
@@ -1,4 +1,4 @@
-﻿namespace FsCheck
+﻿namespace rec FsCheck
```
kurtschelfthout, line 1: Can you explain why the `rec` is necessary?

src/FsCheck/ArbMap.fs
```diff
@@ -11,31 +11,48 @@ type IArbMap =
     abstract ArbFor: Type -> Arbitrary<obj>
     abstract ArbFor<'T> : unit -> Arbitrary<'T>
 
+type private ArbMapInit = 
+| FromTypeClass of TypeClass.TypeClass<Arbitrary<obj>>
+| FromDiscover of (Type * ArbMap option)
```
kurtschelfthout, line 16: This could be two constructors or perhaps factory methods on `ArbMap`?

src/FsCheck/FSharp.ArbMap.fs
```diff
@@ -14,7 +14,7 @@ module ArbMap =
     /// discovered on the given Type. See mergeWith<'TArb> for more info on what the shape of instancesType can be.
     let mergeWithType (instancesType: Type) (existingMap: IArbMap) =
         ArbMap(instancesType, existingMap :?> ArbMap) :> IArbMap
-
+
```
kurtschelfthout, line 17: unnecessary indent

src/FsCheck/FSharp.ArbMap.fs
```diff
@@ -28,7 +28,23 @@ module ArbMap =
     /// parameters of type Arbitrary<'T> to obtain them. Arbitrary instance of the correct type are automatically
     /// injected on construction of a particular type.
     let mergeWith<'TArb> (existingMap: IArbMap) = mergeWithType typeof<'TArb> existingMap
+    
+    /// Return a new Type to Arbitrary map that merges the existing map with the Arbitrary<'T> returned by the given function.
+    let mergeFactory (factory: unit -> Arbitrary<'b>) (existingMap: IArbMap) = 
+        (existingMap :?> ArbMap).MergeFactory(Func<unit,Arbitrary<'b>>(factory)) :> IArbMap
```
kurtschelfthout, line 34: I think `MergeFactory` should go on `IArbMap` instead of the casts.

src/FsCheck/ArbMap.fs
farlee2121: I opted for rec over TypeMapInit taking an IArbMap and assuming a downcast to ArbMap. ArbMap needs the union for the constructor, but the union needs ArbMap for the case type. I think we could reduce the scope of the recursive definition

src/FsCheck/FSharp.ArbMap.fs
farlee2121: Makes sense to me

farlee2121: Well, something to consider real quick. I implemented it this way to mimic mergeWithType. I thought, perhaps, we didn't want to expose any modification on IArbMap. IArbMap is currently a read-only abstraction, and MergeFactory would change that. I'm not sure how much it matters, since all the modification is exposed by proxy...

src/FsCheck/ArbMap.fs
farlee2121: I struggled with this constructor for a good while.

Static factories are out. There must be a reference to the ArbMap being initialized in the typeclass needed for initializing the ArbMap, else the typeclass wont resolve expected factory scenarios.

Here's an example of trying to split into two constructors ([comparison](https://github.com/farlee2121/FsCheck/compare/direct-arb-registration...farlee2121:FsCheck:separate-constructors?expand=1)). It compiles, but it throws 
> The initialization of an object or value resulted in an object or value being accessed recursively before it was fully initialized.

This recursion can't be fixed separating discover from registering the ArbMap instance as a config parameter([example](https://github.com/farlee2121/FsCheck/compare/direct-arb-registration...farlee2121:FsCheck:separate-constructors-2?expand=1)). If the ArbMap is not registered with the discover, then factories that require it will be filtered out by discover and lots of tests fail. 

We could solve this with a mutable field, but that didn't seem like a good option

farlee2121: Short version: There is a tricky mutual dependence between the ArbMap instance and the typeclass it relies on which makes construction difficult

src/FsCheck/ArbMap.fs
farlee2121: Reduced the recursion scope to just the two type definitions

farlee2121: Reduced the recursion scope to just the two type definitions

farlee2121:
@kurtschelfthout It appears .NET 5.0 is no longer available on the build server. Should I update the test project to use .NET 6?

kurtschelfthout:
> @kurtschelfthout It appears .NET 5.0 is no longer available on the build server. Should I update the test project to use .NET 6?

Strange it's already removed, but agree that seems to be the problem. Go ahead, thanks!

farlee2121:
@kurtschelfthout The build is fixed.

PR #617, opened 2022-08-15
Add prop rejection helpers

ComradeVanti:
I sometimes find myself wanting to reject a property, but not being able to just return `false`, because my test returns a `Property`, such as in this case:
```fsharp
match parse s with
| Some value -> value = original |@ "Parsed incorrect value"
| None -> false // Cannot return false, since the other branch returns a Property
```
In this case I would do
```fsharp
| None -> Prop.rejectWith "Could not parse"
```

### DIFF
--- a/src/FsCheck/FSharp.Prop.fs
+++ b/src/FsCheck/FSharp.Prop.fs
@@ -77,6 +77,14 @@ module Prop =
     [<CompiledName("Discard")>]
     let discard() = raise DiscardException
 
+    /// A rejected property
+    [<CompiledName("Rejected")>]
+    let rejected = Prop.ofBool false
+    
+    /// Reject a property with a specific message
+    [<CompiledName("RejectWith")>]
+    let rejectWith message = rejected |> label message
+
 ///Operators for Prop.
 [<AutoOpen>]
 module PropOperators =


ploeh:
Have you considered using an assertion library? I like [Unquote](http://www.swensensoftware.com/unquote/), but I also use the assertion APIs that come with testing frameworks (e.g. [xUnit.net](https://xunit.net)).

An assertion like the above I'd typically write like this with Unquote:

```fsharp
Some original =! s
```

`=!` is an Unquote operator that I usually read as _must equal._

That's going to give you a nice descriptive error message if the assertion fails, _including_ the values of `original` and `s`.

In assertions, there's no reason to pattern-match on an option value, since options have structural equality.

kurtschelfthout:
Thanks for considering a contribution to FsCheck! 

However I can't accept this PR - on the grounds that `false @| "Could not parse"`  is just as short and doesn't need extra API.

See also: https://wiki.haskell.org/Fairbairn_threshold which I don't think is reached here.

ComradeVanti:
@ploeh I have not used unquote, but I also have the `=!` operator in my project (though I called it `=?`). I have not thought of how easily options can be compared without a `match` in the way you showed. Thanks for that.

@kurtschelfthout I though it would be more readable with an explicit `Prop` API, but I can see how that opens the door for many more small and somewhat niche functions to be added to it, so I get leaving it out.

PR #618, opened 2022-10-04
+ Map overload to make it F#+ friendly

gusty:
I propose to add this overload to make it F#+ friendly.

F#+ (as many other libs) defines also `<!>`, `<*>` and `>>=`, but in a generic way and actually it doesn't cause any problem with `<*>` and `>>=`, but for `<!>` it expects a `Map` static member.

This is a small adjustment, backwards compatible and it doesn't cause any problem here.

Eventually if you don't agree to expose a `Map` static member, I can hide it with `[<EditorBrowsable(EditorBrowsableState.Never)>]`

### DIFF
--- a/src/FsCheck/Gen.fs
+++ b/src/FsCheck/Gen.fs
@@ -649,8 +649,11 @@ type Gen<'a> with
     /// Lifted function application = apply f to a, all in the Gen applicative functor.
     static member (<*>) (f, a) = apply f a
 
-    /// Like <*>, but puts f in a Gen first.
+    /// Like <*>, but puts f in a Gen first. An alias for Map.
     static member (<!>) (f, a) = Gen.constant f <*> a
+    
+    /// Like <*>, but puts f in a Gen first.
+    static member Map (f, a) = Gen.constant f <*> a
 
     /// Bind operator; runs the first generator, then feeds the result
     /// to the second generator function.


src/FsCheck/Gen.fs
```diff
@@ -646,12 +646,18 @@ module Gen =
...
 
+    /// Puts a in a Gen.
+    static member Return (a) = Gen.constant a
+    
     /// Lifted function application = apply f to a, all in the Gen applicative functor.
     static member (<*>) (f, a) = apply f a
 
-    /// Like <*>, but puts f in a Gen first.
+    /// Like <*>, but puts f in a Gen first. An alias for Map but with arguments swapped.
     static member (<!>) (f, a) = Gen.constant f <*> a
```
kurtschelfthout, line 656: I know I wrote this, but now that you've attracted my attention to it :)

I think this should just call `Gen.map` because that's what it is. Also the part of the comment `Like <*>, but puts f in a Gen first.` is confusing (again, my bad). I suggest just keeping your `An alias for map with arguments swapped`.

src/FsCheck/Gen.fs
```diff
@@ -646,12 +646,18 @@ module Gen =
...
+    
+    /// Like <*> with arguments swapped, but puts f in a Gen first.
+    static member Map (a, f) = Gen.constant f <*> a
```
kurtschelfthout, line 659: Same here, just use `Gen.map` + doc

kurtschelfthout:
No hard objections, just wondering if some of this won't clash with `Gen.map` on the `Gen` module which has `CompiledName("Map")`? Or maybe F#+ can leverage that?

Also please add a comment to clarify what the intent is of these methods, i.e. to support F#+, as I doubt anyone will remember after some time and they might get removed inadvertently....

gusty:
Hi Kurt, thanks for your feedback.

> No hard objections, just wondering if some of this won't clash with Gen.map on the Gen module which has CompiledName("Map")? Or maybe F#+ can leverage that?

Let me give it a try first, in order to make sure the interaction is as smooth as possible.

src/FsCheck/Gen.fs
gusty: I agree. Done !

src/FsCheck/Gen.fs
gusty: Done !

gusty:
I tried it and it works perfectly.
It's true that before applying this, type inference gets confused and shows an accessibility error mentioning a private `Map` implementation (possibly from `IGen`) which is definitely misleading, since the compiled names are not taken into account for trait-calls, but the F# ones.
It seems that the F# compiler should be adjusted in the regard.

I left `Return` hidden for tooling as probably could be confusing for the user, can do the same with `Map` if you want, as there's already a `.map` function.

kurtschelfthout:
Thanks!

> I left Return hidden for tooling as probably could be confusing for the user, can do the same with Map if you want, as there's already a .map function.

Yes, please.

gusty:
Done !

Issue #633, opened 2023-06-19
Allow asynchronous tests to run on Task<'a>


ploeh:
That FsCheck can run asynchronous tests is one of its major advantages over competitors. It'd be nice, however, if it could run methods/actions that return `Task<'a>` instead of simply `Task`.

Particularly, with the new `task` computation expression, one can easily write an asynchronous (integration) test, but even if such a test body ends with an assertion, the expression has the type `Task<unit>`, rather than `Task`.

One can get around this issue by upcasting the expression with `:> Task`. If, however, one wishes to avoid adding an extra line of code for only that purpose, one runs into [this issue](https://github.com/dotnet/fsharp/issues/15432).

If, currently, one tries to pass a `Task<unit>` to `Prop.forAll` if compiles fine, but fails at run time with this error message:

>     No instances of class FsCheck.Testable+ITestable`1[T] for type
>     System.Threading.Tasks.Task`1[Microsoft.FSharp.Core.Unit]

This is on FsCheck.Xunit 3.0.0-beta2.

Since `Task<'a>` derives from `Task` is seems to me that a fix should be a fairly straightforward.

If you agree that such a fix is desired, I may try my hand at it myself. (I've been away from contributing to FsCheck for years, now, so I hope I'm not biting off more than I can chew.)

kurtschelfthout:
Agree that a fix is desired. Actually not sure why it doesn’t already work! 

Let me know if you get stuck, happy to assist (although it’s quite busy these days, so may take a couple days to reply)

kurtschelfthout:
It's known now, in that I see it too :) 

Afaik this is due to some incompatibility between fake and installed MSBuild versions. This kind of thing unfortunately happens a lot. For a project like FsCheck which is essentially untouched, the amount of churn introduced by .NET, Fake, paket and all the build paraphernelia keeps surprising me. Excuse the rant.

In any case it's moot, because since you want to modify 3.0, you need to check out (and PR to) the `fscheck3` branch. It is substantially different (i.e. includes the async support...whereas master doesn't). 

For that, as well as the eternal  prerelease state of 3.0, there is no one to blame but myself.

In any case I managed to update the `fscheck3` branch to the latest paket and fake, and after pinning to .NET 6 - you'll need a .NET 6 SDK - that builds fine, also in CI: https://ci.appveyor.com/project/kurtschelfthout/fscheck/builds/47399941

Hope this helps!

ploeh:
All looks green now, thank you!

[...]

PR #634, opened 2023-06-27
Enable properties that return Task<'a>

ploeh:
This is intended to address #633.

I'm a little concerned whether the addition of `TaskGeneric` overrides or shadows the more specific `TaskBool`, but all tests pass. Still, I wanted to call attention to this question, since I don't have complete understanding of how the underlying machinery works.


kurtschelfthout (approved):
LGTM, the generic should not pose a problem. Not sure why I didn't do that in the first place.

PR #646, opened 2023-10-07
Generates negative decimal values

sasmithjr:
Heya! When writing some tests with FsCheck yesterday, I discovered the tests taking forever because `Gen.filter (fun d -> d <= 0)` for decimals was never satisfying the predicate.

I copied how negative values are created for `NormalFloat`.

I added a test that uses the above filter to make sure negative values are generated. If there's a regression that reverts back to previous behavior (i.e. only generating positive decimals), this test will "fail" by hitting a timeout in CI. Is that okay?

Thank you for the great library, and have a great weekend!


kurtschelfthout:
LGTM, thanks!

PR #665, opened 2024-02-25
Fix release

Smaug123:
The bug was not setting a user-agent, but this way we are also more explicit if things go wrong.

> Requests with no User-Agent header will be rejected. If you provide an invalid User-Agent header, you will receive a 403 Forbidden response.
(https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api?apiVersion=2022-11-28#user-agent)

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -530,7 +530,7 @@ type GitHubReleaseResponse =
         [<JsonPropertyName "name">]
         Name : string
         [<JsonPropertyName "published_at">]
-        PublishedAt : DateTime
+        PublishedAt : System.Nullable<DateTime>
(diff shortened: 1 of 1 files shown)


build.fsx
```diff
@@ -530,7 +530,7 @@ type GitHubReleaseResponse =
         [<JsonPropertyName "name">]
         Name : string
         [<JsonPropertyName "published_at">]
-        PublishedAt : DateTime
+        PublishedAt : System.Nullable<DateTime>
```
Smaug123, line 533: (I observed this to be null in my testing)

Issue #664, opened 2024-02-25
Packaging improvements


kurtschelfthout:
With the recent changes in deploy, I believe we are in a good position to modernize FsCheck packaging.

- [ ] Enable source linking
- [ ] Enable deterministic build
- [ ] Check if we can include compiler flags (not sure fsc does that)

Imo source linking is pretty useful. It used to be enabled but it bitrotted, I think it's much more standardized now.

Goal would be to have a squeky clean Nuget package explorer health check. Currently looks like this:

![image](https://github.com/fscheck/FsCheck/assets/164917/24e74e27-a432-4ac8-9592-49d07001f4eb)

Smaug123:
I'm looking at sourcelink now. (I need to get round to it for my own packages anyway.)

Smaug123:
To my great surprise, [there is no public way of reading SourceLink metadata](https://developercommunity.visualstudio.com/t/Is-there-an-API-for-reading-Source-Link-/1468642)?! (Maybe this is not to my surprise, because this is MsBuild we're talking about.

Smaug123:
I'm not convinced that EmbedUntrackedSources is what we want (at least, I'm not convinced until I can know what those untracked sources *are*).

Smaug123:
Before working out that mystery, I've pulled out https://github.com/fscheck/FsCheck/pull/670 for just the deterministic builds bit.

Smaug123:
The compilation options thing is not supported in F# (https://github.com/dotnet/fsharp/issues/12002).

Smaug123:
To summarise:

* Deterministic builds and SourceLink are now working (as you noted in https://github.com/fscheck/FsCheck/pull/670#issuecomment-1963022094), modulo [a bug in the F# compiler](https://github.com/dotnet/fsharp/issues/16768) and [a bug in the interaction between MsBuild and the compiler](https://github.com/dotnet/fsharp/issues/16771). 
* Compilation options [are not supported](https://github.com/dotnet/fsharp/issues/12002).

I think this ticket can be closed, because it's as complete as it's practical to be.

PR #670, opened 2024-02-25
Deterministic builds

Smaug123:
Ref: #664

Works for me in the sense that the resulting nupkg lets me "go to definition" in Rider and I see the F# source. I'll do the same check with the nupkgs that AppVeyor gives me, just to make sure.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,6 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
+    <Deterministic>true</Deterministic>
     <Copyright>Copyright 2008-2024</Copyright>
     <Authors>Kurt Schelfthout and contributors</Authors>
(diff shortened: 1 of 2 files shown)


Smaug123:
Hmm: https://github.com/dotnet/sourcelink/blob/a22426a46745b8fe77b9d306fa4bc1d03eff44ea/README.md

> If your project uses .NET SDK 8+ and is hosted by the above providers it does not need to reference any Source Link packages or set any build properties.

Smaug123:
Yeah, I seem to get that behaviour even without a reference to SourceLink, so I think this should Just Work (TM).

Smaug123:
It wouldn't be the `dotnet` CLI without nondeterministic and inconsistent behaviour - there's a solid chance this just doesn't work (https://github.com/dotnet/sdk/issues/16325#issuecomment-916297293 for example) and we may simply have to wait for an SDK bump after this goes in.

kurtschelfthout:
Progress:

image

PR #672, opened 2024-02-25
Fix release notes

Smaug123:
MsBuild is kind of hilarious really. If there's a better way to achieve this, I have been unable to find it.

I've verified this myself locally up to and including checking that the nuspec file contains a `<releaseNotes>` which looks reasonable.


kurtschelfthout:
Good old base64. Saving people from escape codes since 2006.

Issue #675, opened 2024-03-05
Tests not ending with custom shrinker


rynoV:
I'm running into a case where I have a custom shrinker which calls two other shrinkers (one from FsCheck and one custom), and normally the tests using this arbitrary instance pass (so I would expect the custom shrinking code not to be called). However it seems like when I add code which evaluates the sequence of shrinks generated from either the FsCheck shrinker or my custom shrinker, the tests stall.

For example just adding the line

```fsharp
let _ = (ArbMap.defaults |> ArbMap.arbitrary |> Arb.toShrink) dates |> Seq.length
```

(`dates: DateTime[]`) in my shrinker definition stalls the tests, and after commenting that out they pass.

I'm guessing this is because the shrunk sequence might be infinite, but I couldn't find any docs on this. Also, does the shrinker get evaluated even when the tests don't fail?

I'm on version 3.0.0-rc3

kurtschelfthout:
FsCheck has no magic way to prevent shrinking if you literally evaluate it explicitly.

Shrinking can take a long time. It can also get stuck in a loop, if shrinking a value yields the same value again somewhere along the way.

rynoV:
Sorry maybe I didn't explain well. I'm defining an arbitrary instance like

```fsharp
let arbDataSeriesItemTest (config: DataSeriesTestConfig) =
    { new Arbitrary<...>() with
        override x.Generator = ...

        override x.Shrinker(dates) = ...
        }
```

and the issue I'm having is that putting

```fsharp
let _ = (ArbMap.defaults |> ArbMap.arbitrary |> Arb.toShrink) dates |> Seq.length
```

inside my `Shrinker` method override freezes my tests, even when they're passing, in which case I would expect `Shrinker` to never be called.

I suppose it doesn't really matter, since the implementation shouldn't evaluate the infinite sequence anyways, it was just confusing behaviour and took me a while to figure out what was happening.

I'll make a PR adding to the docs that the sequence of shrinks may be infinite, and evaluating that sequence may freeze the tests even if no tests are failing.

kurtschelfthout:
I see. Yes, FsCheck relies on the laziness of the returned sequence. It will call `Shrinker` during normal evaluation of the test, and capture the resulting sequence. This allows it to pick up shrinking with the values the test failed with. It effectively captures `Shrinker(dates)` in a closure, but only evaluates the resulting sequence one at a time.

rynoV:
[...]

PR #676, opened 2024-03-05
Update docs on infinite data in shrinkers

rynoV:
#675


docs/TestData.fsx
```diff
@@ -35,6 +35,13 @@ This dictionary is used to find an arbitrary instance for properties that have a
 based on the argument's type.
 
 Arbitrary instances have some helper functions in `Arb`.
+
+## Shrinkers
+
+When writing custom shrinkers, keep in mind that:
+
+- The sequence of shrinks generated by the built-in shrinkers may be infinite.
```
kurtschelfthout, line 43: This is not accurate. The built-in shrinkers all terminate for all inputs. If they do not, it's a bug. 

Shrinking may take a long time though, but typically in pathological, constructed cases (such as a test that only fails for a particularly long input and for nothing else, and so all shrinks are tried without success)

A common problem with writing custom shrinkers is that they fail to terminate, typically because a shrinker returns a value that was already tried.

docs/TestData.fsx
```diff
@@ -35,6 +35,13 @@ This dictionary is used to find an arbitrary instance for properties that have a
...
+- Your custom shrinker function may be called even if no tests fail, so if your tests start stalling remember to also check your shrinker for infinite loops
```
kurtschelfthout, line 44: see my comment on this on the corresponding issue for clarification

PR #681, opened 2024-05-30
Bugfix FsCheck.Fluent.Gen.Fresh

Andreas-Dorfer:
I might be missing something, but it seems to me like the `create` parameter of `FsCheck.Fluent.Gen.Fresh` has the wrong type. It should be `Func<'T>` (not `Func<Gen<'T>>`).

For comparison, the `create` parameter of `FsCheck.FSharp.Gen.fresh` is of type `unit -> 'T`.


kurtschelfthout:
Indeed! Thanks for fixing.

PR #685, opened 2024-11-26
NUnit4

### LINKED ISSUE #650: feat(FsCheck.NUnit): Support NUnit >= 4.0.0
Today FsCheck.NUnit relies on a `< 4.0.0` version of NUnit.

DragonSA:
This pull request extends the dependency requirement for FsCheck.NUnit to also include NUnit 4.  

I've taken a stab at updating FsCheck to NUnit 4 (`dotnet paket update nunit`).  The one issue is NUnit 4 no longer supports `netstandard2.0`, but instead targets `net6` (and it seems Paket/FsCheck does not support targeting .NET Framework).  Other than that there were no changes needed (i.e. no changes to `FsCheckPropertyAttribute.fs`).  

As others have reported, and I can confirm, just forcing FsCheck.NUnit v2 to use NUnit 4 works.  

I thought this approach was the most versatile (and does not break backward compatibility).  

Fixes #650.  

Sponsored by [Counterpoint Dynamics](https://cpdynamics.co.za).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,9 @@ project.lock.json
 .fsdocs/
 output/
 tmp/
+
+#######
+# Rider
+#######
+.idea/
+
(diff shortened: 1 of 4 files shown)


kurtschelfthout:
Hey, thanks for this - however, you can just contribute to master, the fscheck3 branch has been merged, and all fecheck 3 releases are now from master! In other words, this PR as stated won't do anything unfortunately....sorry for the confusion. I will delete the fscheck3 branch to avoid this problem in future.

Finally, please take into account FsCheck no longer uses paket. We just use `dotnet pack` now. To be honest, I don't know how to achieve what you did here with the paket files in vanilla `dotnet pack`...

DragonSA:
Thanks for the direction.  I've rebased onto master and created #691.

PR #691, opened 2024-12-31
NUnit4

### LINKED ISSUE #650: feat(FsCheck.NUnit): Support NUnit >= 4.0.0

DragonSA:
This pull request extends the dependency requirement for FsCheck.NUnit to also include NUnit 4.

I've taken a stab at updating FsCheck to NUnit 4 (dotnet paket update nunit). The one issue is NUnit 4 no longer supports netstandard2.0, but instead targets net6 (and it seems Paket/FsCheck does not support targeting .NET Framework). Other than that there were no changes needed (i.e. no changes to FsCheckPropertyAttribute.fs).

As others have reported, and I can confirm, just forcing FsCheck.NUnit v2 to use NUnit 4 works.

I thought this approach was the most versatile (and does not break backward compatibility).

Fixes #650, related to #685.  

Sponsored by [Counterpoint Dynamics](https://cpdynamics.co.za/).


kurtschelfthout:
Great, thanks!

Issue #684, opened 2024-10-07
`bool` ≈ `Task<bool>`, `Property` ≉ `Task<Property>`


brianrourkeboll:
The behavior added in #634 to address #633 feels confusing to me when contrasted with the behavior of tests that return `Task<bool>`, especially for xUnit/NUnit tests annotated with `PropertyAttribute`.

That is:

1. Tests marked with `PropertyAttribute` that return `bool` _fail_ when the `bool` is `false`.
2. Tests marked with `PropertyAttribute` that return `Task<bool>` _fail_ when the `bool` is `false`.
3. Tests marked with `PropertyAttribute` that return `Property` _fail_ when the `Property` is falsy.
4. Tests marked with `PropertyAttribute` that return `Task<Property>` _pass_ when the `Property` is falsy.

I would expect 4 to fail just as 1–3 do.

```csharp
// Fails as expected.
[Property]
public bool Test(bool b) =>
    b && !b;

// Fails as expected.
[Property]
public Property Test(bool b) =>
    Prop.Label(b && !b, "Some label.");

// Fails as expected.
[Property]
public Task<bool> Test(bool b) =>
    Task.FromResult(b && !b);

// Passes. I would expect this to fail.
[Property]
public Task<Property> Test(bool b) =>
    Task.FromResult(Prop.Label(b && !b, "Some label."));

// Passes. I would expect this to fail.
[Property]
public Task<Property> Test(bool b) =>
    Task.FromResult(Prop.ToProperty(b).And(!b));
```

```fsharp
// Fails as expected.
[<Property>]
let test1 b =
    b && not b

// Fails as expected.
[<Property>]
let test2 b =
    (b && not b) |> Prop.label "Some label."

// Fails as expected.
[<Property>]
let test3 b =
    Task.FromResult (b && not b)

// Passes. I would expect this to fail.
[<Property>]
let test4 b =
    Task.FromResult ((b && not b) |> Prop.label "Some label.")

// Passes. I would expect this to fail.
[<Property>]
let test5 b =
    Task.FromResult (b .&. not b)
```

I understand the original desire not to need to explicitly upcast `Task<unit>` to `Task`, but the difference in treatment between `Task<bool>` and `Task<Property>` is dangerous, because it is far too easy to write a test that returns `Task<Property>` and have it pass for the wrong reason.

[...]

PR #693, opened 2025-01-16
#684 Task<Property> - Draft

alex-bogomaz:
Hello, 

this draft PR is for discussion regarding #684.

Why not remove ``ResultContainer`` altogether and force Task result on the level where Property is created out of Testable?
Do you see any issues with the approach?

### DIFF
--- a/src/FsCheck/Testable.fs
+++ b/src/FsCheck/Testable.fs
@@ -39,18 +39,8 @@ type Result =
         | (_,Outcome.Passed) -> r
         | (Outcome.Rejected,Outcome.Rejected) -> l //or r, whatever
 
-type ResultContainer = 
-    | Value of Result
-    | Future of Task<Result>
-    static member private MapResult2(f, l, r) =
-        match (l,r) with
-        | (Value vl,Value vr) -> f (vl, vr) |> Value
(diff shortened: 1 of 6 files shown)


src/FsCheck/Testable.fs
```diff
@@ -103,35 +93,51 @@ module private Testable =
...
+
+            match (b.IsCanceled, b.IsFaulted) with
+                | (false,false) -> Res.ofBool b.Result
+                | (_,true) -> Res.failedException b.Exception
+                | (true,_) -> Res.failedCancelled
             |> ofResult
+
+        let ofTaskProperty (b:Task<Property>) :Property = 
+            try 
+                b.Wait()
```
brianrourkeboll, line 119: > Why not remove `ResultContainer` altogether and force Task result on the level where Property is created out of Testable?
> Do you see any issues with the approach?

I think this call to `Task.Wait()` would be problematic.

kurtschelfthout: Yes agreed - it looks like this change would (at least partly) break parallel and/or async execution, Such changes are unfortunately hard to test, but the original contribution of parallel testing did verify actual speedups on non-trivial tests.

PR #694, opened 2025-01-20
Support `Async<'Testable>` & `Task<'Testable>`

### LINKED ISSUE #684: `bool` ≈ `Task<bool>`, `Property` ≉ `Task<Property>`

brianrourkeboll:
@kurtschelfthout This is basically just the code from [my comment here](https://github.com/fscheck/FsCheck/issues/684#issuecomment-2558624284). Upon further thought and experimentation, I think it might actually be valid, if not  terribly elegant—but perhaps there is something that I still do not understand and that the tests do not cover. It essentially hinges on whether what I am doing with [`Shrink.getValue` inside of `Gen.promote` (after applying `runner`)](https://github.com/fscheck/FsCheck/blob/441f565d320b51a503e8a3fbdd603b33386a34d7/src/FsCheck/Testable.fs#L125-L143) is valid in all scenarios.

Fixes #684.[^1]

- Add support for `Async<'Testable>` and `Task<'Testable>` as first-class testables.

  That is, `Async<'Testable>` and `Task<'Testable>` are now valid testables for any valid instantiation of `'Testable`, just as `Lazy<'Testable>`, `Gen<'Testable>`, or `'Generatable -> 'Testable` already were.

  ```fsharp
  [<Property>]
  let ``Unit - used to pass, still passes`` () = ()

  [<Property>]
  let ``Task<unit> - used to pass, still passes`` () = task { return () }

  [<Property>]
  let ``Task - used to pass, still passes`` () = task { return () } :> Task

  [<Property>]
  let ``Async<unit> - used to pass, still passes`` () = async { return () }

  [<Property>]
  let ``Property - used to fail, still fails`` b = (b && not b) |> Prop.label "b && not b"

  [<Property>]
  let ``Task<Property> - used to pass, now fails`` b = task { return (b && not b) |> Prop.label "b && not b" }

  [<Property>]
  let ``Async<Property> - used to pass, now fails`` b = async { return (b && not b) |> Prop.label "b && not b" }

  [<Property>]
  let ``Task<Lazy<bool>> - used to pass, now fails`` b = task { return lazy b }

  // Etc.
  ```

[^1]: Note that this is technically a breaking change in that it breaks the example added in #634. An explicit upcast to non-generic `Task` (`… :> Task`) will now be needed for tests involving `Task<'T>` where `'T` is not itself a testable type to pass. This matches the behavior for the unwrapped equivalent:

    ```fsharp
    [<Property>]
    let ``An int is not a testable`` (i : int) = i
    // System.Exception: No instances of class FsCheck.Testable+ITestable`1[T] for type System.Int32
    
    [<Property>]
    let ``An Async<int> is now also not a testable`` (i : int) = async { return i }
    // System.Exception: No instances of class FsCheck.Testable+ITestable`1[T] for type System.Int32
    
    [<Property>]
    let ``A Task<int> is now also not a testable`` (i : int) = task { return i }
    // System.Exception: No instances of class FsCheck.Testable+ITestable`1[T] for type System.Int32
    ```
    
    The behavior originally desired in #633 for `Task<unit>` is, however, kept: an upcast from `Task<unit>` to `Task` is _not_ required, since `unit` is a testable type.

### DIFF
--- a/examples/FsCheck.Examples/Examples.fs
+++ b/examples/FsCheck.Examples/Examples.fs
@@ -194,7 +194,7 @@ Check.One(bigSize,fun (s:Simple) -> match s with Leaf2 _ -> false | Void3 -> fal
 
 Check.One(bigSize,fun i -> (-10 < i && i < 0) || (0 < i) && (i < 10 ))
 Check.Quick (fun opt -> match opt with None -> false | Some b  -> b  )
-Check.Quick (fun opt -> match opt with Some n when n<0 -> false | Some n when n >= 0 -> true | _ -> true )
+Check.Quick (fun opt -> match opt with Some n when n < 0 -> false | Some n when n >= 0 -> true | _ -> true )
 
 let prop_RevId' (xs:list<int>) (x:int) = if (xs.Length > 2) && (x >10) then false else true
 Check.Quick prop_RevId'
--- a/examples/FsCheck.NUnit.CSharpExamples/SyncVersusAsyncExamples.cs
+++ b/examples/FsCheck.NUnit.CSharpExamples/SyncVersusAsyncExamples.cs
@@ -1,0 +1,56 @@
+﻿using System.Threading.Tasks;
+using FsCheck.Fluent;
+using NUnit.Framework;
+
+namespace FsCheck.NUnit.CSharpExamples;
+
+public class SyncVersusAsyncExamples
+{
+    [Property]
+    public Property Property_ShouldPass(bool b)
+    {
+        return (b ^ !b).Label("b ^ !b");
+    }
+
+    [Property]
+    public Property Property_ShouldFail(bool b)
+    {
+        return (b && !b).Label("b && !b");
+    }
+
+    [Property]
+    public async Task Task_ShouldPass(bool b)
+    {
+        await DoSomethingAsync();
+        Assert.That(b ^ !b);
+    }
+
+    [Property]
+    public async Task Task_Exception_ShouldFail(bool b)
+    {
+        await DoSomethingAsync();
+        Assert.That(b && !b);
+    }
+
+    [Property]
+    public async Task Task_Cancelled_ShouldFail(bool b)
+    {
+        await Task.Run(() => Assert.That(b ^ !b), new System.Threading.CancellationToken(canceled: true));
+    }
+
+    [Property]
+    public async Task<Property> TaskProperty_ShouldPass(bool b)
+    {
+        await DoSomethingAsync();
+        return (b ^ !b).Label("b ^ !b");
+    }
+
+    [Property]
+    public async Task<Property> TaskProperty_ShouldFail(bool b)
+    {
+        await DoSomethingAsync();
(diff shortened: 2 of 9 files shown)


kurtschelfthout:
This is looking good, I'll need a bit more time to properly review  - more of a weekend thing.

Meanwhile thanks for the detailed notes, very helpful.

kurtschelfthout:
LGTM. Great work, thanks!

Issue #696, opened 2025-01-23
How to use last step's seed to reproduce the failing case


AT-WH:
### Background

I've created a simple test that's meant to fail at some point

```fsharp
    [<Property>]
    let test(input: byte) =
        // doomed to fail
        Assert.NotEqual(input, 13uy)
```

For a run I received the following message:

```
FsCheck.Xunit.PropertyFailedException : 
Falsifiable, after 26 tests (0 shrinks) (6597590333766474912,8079018595519746725)
Last step was invoked with size of 27 and seed of (11697003081037599755,2368185113887541527):
Original:
   ...
```

When I copy `6597590333766474912,8079018595519746725` and inject it into `Property` attribute's constructor, I get the same result everytime.

```fsharp
    [<Property(Replay="(6597590333766474912,8079018595519746725)")>]
    let test(input: byte) =
        // fails in a repeatable way
        Assert.NotEqual(input, 13uy)
```

This is all as expected so far.

### Problem

How shall I use the last step's seed to jump directly to the failing case?

I've tried naively this, but it didn't work:
```fsharp
    [<Property(Replay="(11697003081037599755,2368185113887541527)")>]
    let test(input: byte) =
        // actually, this passes all the time :)
        Assert.NotEqual(input, 13uy)
```

I've also tried using the size of 27, but that didn't work either.

```fsharp
[<Property(Replay="(11697003081037599755,2368185113887541527)", StartSize=27)>]
```

ploeh:
I don't know if there's an easier way, but if you want to debug, you may be able to set a breakpoint in Visual Studio that first triggers the 27th time it's hit...

kurtschelfthout:
You can pass the size as part of the Replay string - like https://github.com/fscheck/FsCheck/blob/b4a0047ac66329d632ea919b0328e1ed3e27da27/examples/FsCheck.XUnit.CSharpExamples/ReplayExample.cs#L25

We should make this clear in the message somehow and/or add to the docs around here https://fscheck.github.io/FsCheck//TipsAndTricks.html#Replaying-a-failed-test

AT-WH:
Hi Kurt,

Once again, thanks for a very accurate and helpful response.
Not sure how I could help with the docs, but I tried adding this information to the Property attribute's IntelliSense comment:
https://github.com/fscheck/FsCheck/pull/697

What do you think?

PR #697, opened 2025-01-24
Adding more information to `Property` attribute's doc comment

AT-WH:
This is a small change to add more information as discussed in https://github.com/fscheck/FsCheck/issues/696.


kurtschelfthout:
great, thanks!

Issue #690, opened 2024-12-27
upgrade FsCheck.Xunit and FsCheck.NUnit to new major versions


draptik:
Both Xunit and NUnit have had major releases which are not reflected in FsCheck.Xunit and FsCheck.NUnit.

- [ ]  Xunit: new major release v3 - needs creation of new assembly + package FsCheck.Xunit3 
- [x]  NUnit: new major release v4

The corresponding FsCheck implementations might be updatable?

Example:

![image](https://github.com/user-attachments/assets/6490a51e-38bf-41a7-8cc4-8f6563526b0f)

ploeh:
Do the new versions of those two libraries include breaking changes? If they use semantic versioning, I'd expect them to, but haven't investigated.

If so, do the breaking changes affect FsCheck.Xunit or FsCheck.NUnit? If so, we need to investigate how to resolve such issues. If not, it's possible that we could relax the version restrictions to allow the new major versions...

To be clear, I don't have the time to do that at the moment.

draptik:
I just wanted to inform...

I am just learning how to use this library.

This might not be an issue for other people...

@ploeh feel free to close this ticket

tomachristian:
This is definitely holding us from upgrading to XUnit V3, which does have breaking changes. We tried recompiling FsCheck.XUnit ourselves targeting XUnit V3, but it does not compile as is.

kurtschelfthout:
I did NUnit, but xunit seems to have moved on to a different nuget package xunit.v3 so it's a bigger lift.

ploeh:
FWIW, in [AutoFixture](https://github.com/AutoFixture/AutoFixture), we ran into a similar problem when xUnit.net revved from v1 to v2. There were breaking changes that we couldn't get around, so we left the original [AutoFixture.Xunit](http://www.nuget.org/packages/AutoFixture.Xunit) on xUnit.net v1 and created a new [AutoFixture.Xunit2](http://www.nuget.org/packages/AutoFixture.Xunit2) package for v2.

tomachristian:
@ploeh Other libraries (`Verify`) did the same thing, they followed the (unfortunate?) convention that XUnit set.

kurtschelfthout:
Yes, it seems they left us no other choice. It does have the advantage it's easy to maintain both v2 and v3 in parallel.

JohSand:
I have taken a stab at an implementation. It seems to work, but I only replicated the test-cases that already existed, and more might be warranted.

PR #710, opened 2025-05-28
Add separate support for xUnit v3

JohSand:
Add simple tests

Implement xUnit testcases for xUnit.v3

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -37,6 +37,8 @@
     <PackageVersion Include="xunit.core" Version="[2.4.1, 3.0.0)" />
     <PackageVersion Include="xunit.extensibility.core" Version="[2.4.1, 3.0.0)" />
     <PackageVersion Include="xunit.extensibility.execution" Version="[2.4.1, 3.0.0)" />
+    <PackageVersion Include="xunit.v3" Version="2.0.2" />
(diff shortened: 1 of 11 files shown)


JohSand:
fixes xUnit part of #690

The changes in xUnit are breaking, and they choose to publish a package with a different name (xUnit.v3). I believe the most reasonable way for this project is to have different nugets for FsCheck.xUnit and FsCheck.xUnit.v3.

In this I attempt the simplest possible translation of the previous implementation to the new abstractions offered in v3. It is mostly copied directly from the v2 implementation.  

It suffers a bit from targeting netstandard2.0 (no task computation expression support), but I feel it is reasonable.

kurtschelfthout:
This looks great - thanks

> It suffers a bit from targeting netstandard2.0 (no task computation expression support), but I feel it is reasonable.

Fwiw fine to target something more recent. E.g. FsCheck.NUnit now targets net6.0 since NUnit's latest major version update.

I'll leave this here for a bit in case you want to change that, but also happy to merge as-is.

JohSand:
Thanks for the review. I believe we can merge as-is. But it occurs to me that I made no changes to build.fsx, so no new Nuget will be published.

kurtschelfthout:
That's ok, shouldn't be too hard to add.

PR #722, opened 2025-12-16
Embed Untracked Sources

martincostello:
Set `EmbedUntrackedSources=true` so that compiler-generated files are included.

I noticed a warning when looking at the package in [NuGet Package Explorer](https://nuget.info/packages/FsCheck/3.3.2):

image

I'm not sure why the two errors are there as `ContinuousIntegrationBuild` and `Deterministic` are both set to `true`:

https://github.com/fscheck/FsCheck/blob/7df8a3c8ac35b3ef1213033fd63d1578dada7335/Directory.Build.props#L9

https://github.com/fscheck/FsCheck/blob/7df8a3c8ac35b3ef1213033fd63d1578dada7335/Directory.Build.props#L5-L7

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,6 +7,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <Deterministic>true</Deterministic>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <Copyright>Copyright 2008-2024</Copyright>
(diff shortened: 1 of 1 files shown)


kurtschelfthout:
I don't know why those warnings either...I seem to vaguely remember one of them, possibly the deterministic one, was related to an F# compiler issue.

Anyway, thanks!

PR #723, opened 2025-12-17
Update fsdocs tool

### LINKED ISSUE #721: fsdocs build failing
(after update to net8.0 sdk?)

fsdocs tool itself could use an update.

alex-bogomaz:
Hello, 

This should fix #721.

Summary of changes: 
-  use "fsdocs-tool" instead of deprecated "fsharp.formatting.commandtool"
-  generate reference documentation for "FsCheck.Xunit" and "FsCheck.NUnit", not just "FsCheck" itself
- update references in "RunningTests.fsx"
- "template.html" : "fsdocs-collection-name-link" is no longer supported, so use "fsdocs-package-project-url" instead
- "template.html" : add additional style for tooltips. Now "FSharp.Formatting" is generating multi-line tooltips without inserting `<br />` tag between lines.  As a result, tooltips that describe types with multiple members are hard to read. Had to fix this using CSS.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -2,8 +2,8 @@
   "version": 1,
   "isRoot": true,
   "tools": {
-    "fsharp.formatting.commandtool": {
-      "version": "9.0.4",
+    "fsdocs-tool": {
+      "version": "21.0.0",
       "commands": [
         "fsdocs"
       ]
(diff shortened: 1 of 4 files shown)


kurtschelfthout:
Good stuff, thanks!

kurtschelfthout:
Took a while but I just used the updated version to publish docs - looks like it broke a number of things...any ideas how to fix?

kurtschelfthout:
Never mind - just needed to delete the old _template.html file. All good now!

